PR #8, opened 2016-07-15
Allowing extending `MongoRelation`

pfcoperez:
With this PR `MongoRelation` becomes a regular class.

By making it mix `Serializable` and providing a factory and a extractor at its companion object we get the same behaviour as it were a case class with the advantage of being extensible.

Note that this PR is not only motivated by  the reuse of ` MongoRelation`  methods  functionality but also by the need of sub-typing when working with low-level extensions of Apache Spark.

### DIFF
--- a/src/main/scala/com/mongodb/spark/sql/MongoRelation.scala
+++ b/src/main/scala/com/mongodb/spark/sql/MongoRelation.scala
@@ -27,9 +27,20 @@ import com.mongodb.spark.rdd.MongoRDD
 import com.mongodb.spark.sql.MapFunctions.documentToRow
 import com.mongodb.spark.sql.MongoRelationHelper.createPipeline
 
-case class MongoRelation(mongoRDD: MongoRDD[BsonDocument], _schema: Option[StructType])(@transient val sqlContext: SQLContext)
+object MongoRelation {
+
+  def apply(mongoRDD: MongoRDD[BsonDocument], _schema: Option[StructType])(sqlContext: SQLContext): MongoRelation =
+    new MongoRelation(mongoRDD, _schema, sqlContext)
+
+  def unapply(arg: MongoRelation): Option[(MongoRDD[BsonDocument], Option[StructType])] =
+    Some((arg.mongoRDD, arg._schema))
+
+}
+
+class MongoRelation(val mongoRDD: MongoRDD[BsonDocument], val _schema: Option[StructType], @transient val sqlContext: SQLContext)
     extends BaseRelation
     with PrunedFilteredScan
+    with Serializable
     with LoggingTrait {
 
   override lazy val schema: StructType = _schema.getOrElse(MongoInferSchema(sqlContext.sparkContext))


rozza:
Can you explain your motivations for wanting to extend `MongoRelation`?

pfcoperez:
I'll give you an example: Sometimes is possible to  perform an scan without going through Spark, you could create a sub-class of your relation aware of that and override `buildScan` so it could try to execute the passed plan in some way - out of Spark - and fallback to the default behaviour otherwise.

https://github.com/Stratio/Crossdata/blob/d6827062364fc10360f65b9f2ce54d0fbe7cfc79/mongodb/src/main/scala/com/stratio/crossdata/connector/mongodb/MongodbXDRelation.scala

In this case, this allow us - at Stratio - to seamlessly boost some Spark SQL queries but I imagine many other people might want to expand the relation functionality for other purposes.

I've found myself opening PRs, and these being accepted, to the Spark-Cassandra connector team  just because I wan't to expanse some functionality. You might want to kick the tires and start developing and using this features without adding it to the connector code in favor of loosely coupled code or to avoid very specific to be embedded in your connector.

rozza:
Thanks for responding @pfcoperez, it sounds an interesting approach. I'd be interested in a concrete scenario where a passed plan is executed out of Spark, so I could better understand what you are trying to achieve.

On the whole I think it would be better if there were some concrete trait in the `BaseRelation` that should be implemented to achieve this goal - similar to `PrunedFilteredScan`. Then library authors can choose to implement it themselves. Have you lobbied for some such mechanism? I'm not sure if the `CatalystScan` trait would be the same thing or not.

Having reviewed the issue, I've come to the conclusion that`MongoRelation` being public was an oversight on my part. It's an implementation detail and as such isn't designed to be extended by end users.  `DefaultSource` should be returning `BaseRelation` for the overridden methods but that ship has sailed with 1.0.  I've added [SPARK-67](https://jira.mongodb.org/browse/SPARK-67) to fix this in 2.0 and it will bring our approach inline with other popular connectors eg: Cassandra and Elastic Search.

As `MongoRelation` is an implementation detail using a case class is a perfect fit for it. However, I can understand you think it hinders your requirements by being one (I'll show that it doesn't later).  I noticed that you don't extend a Cassandra relation in the `CassandraXDSourceRelation` implementation at all and in `ElasticsearchXDRelation` you do extend the `ElasticsearchRelation` which is a private case class (this in itself seems risky).

Do you really need to extend these relation classes at all? Wouldn't it be better to fallback to a specific library's `DefaultSource` implementation in your customised `DefaultSource` implementations? This seems more practical, as it means you don't have to extend internal or private classes that are subject to changes outside your control. Also, library authors may and will differ in their implementations of connectors, so it allows for these differences. That is just an observation without knowing the full domain of the problems your library is trying to solve, so it may not be applicable / desirable.

Back to the issue at hand.  I was able to extend `MongoRelation` much in the same way you do with `ElasticsearchXDRelation` without any code changes:

```
class MongoRelationAlt(mongoRDD: MongoRDD[BsonDocument], _schema: Option[StructType], @transient override val sqlContext: SQLContext)
  extends MongoRelation(mongoRDD: MongoRDD[BsonDocument], _schema: Option[StructType])(sqlContext: SQLContext) with LoggingTrait {
  // Custom implementation details here..
}
```

pfcoperez:
@rozza Your analysis is astoundingly thorough, thanks for that. Other approaches could be tried and, as I am already on holidays and won't have internet connection starting tomorrow, I'll leave my mates to continue the discussion, applying this change would require almost no changes in those libraries currently build on Stratio's connector however.

Although it is possible to extend a case class, it can't ever be extended by other case class. This not only applies for the first subclass but for any possible successor. There are great reasons for this limitation: Problems with provided methods in case classes, mainly.
Making a potentially extensible class a case class is usually regarded as a anti pattern and I'd like to avoid that.

Other issues is whether or not you regard this class as potentially extensible.

Thanks again for your analysis and taking into account the PR.

rozza:
Hi @pfcoperez, I hope you had a good holiday! 

I never got any further feedback or answers to my questions regarding this. I'm currently working on 1.1.0 and 2.0.0 versions of the Spark connector and I would like to come to an amicable conclusion regarding this issue before they ship!

pfcoperez:
Hi @rozza, I am sorry you didn't get further news, I thought you got them through my workmates' conversations with Brian.

We still believe that keeping `MongoRelation` as a regular class would favor its extensibility, at least, as long as there is a better mechanism within the Spark ecosystem. However, as you pointed out, it is not critical.

What is really important for us is that you keep it public or package private (`spark.sql` or `spark` levels) in a way that we can still extend it.

Thank for your time and consideration.

PR #10, opened 2016-10-11
Removed ConflictType as an end type

rozza:
Conflicting types are down converted to a String.

SPARK-84


rozza:
Any feedback before I merge?

evanchooly:
+1

PR #12, opened 2016-12-12
Fix to https://jira.mongodb.org/browse/SPARK-103.

esurijon:
Add a default value of 1 for an optional collstat result key, so partitioner will fallback to only one partition when collection is empty


rozza:
Fixed in rozza/mongo-spark#27

PR #14, opened 2017-01-12
DOCS-9403: update python example for Spark 2

steveren:


examples/src/test/python/introduction.py
```diff
@@ -21,37 +21,28 @@
...
 
     # Save some data
-    charactersRdd = sc.parallelize([("Bilbo Baggins",  50), ("Gandalf", 1000), ("Thorin", 195), ("Balin", 178), ("Kili", 77),
-                                    ("Dwalin", 169), ("Oin", 167), ("Gloin", 158), ("Fili", 82), ("Bombur", None)])
-    characters = sqlContext.createDataFrame(charactersRdd, ["name", "age"])
+    characters = spark.createDataFrame([("Bilbo Baggins",  50), ("Gandalf", 1000), ("Thorin", 195), ("Balin", 178), ("Kili", 77), ("Dwalin", 169), ("Oin", 167), ("Gloin", 158), ("Fili", 82), ("Bombur", None)], ["name", "age"])
     characters.write.format("com.mongodb.spark.sql").mode("overwrite").save()
 
-    # Load the data
-    df = sqlContext.read.format("com.mongodb.spark.sql").load()
```
rozza: We shouldn't delete this as its an example of  loading data from MongoDB.

Otherwise we are just running a SQL query on the local dataframe.

steveren: Hi Ross -- line 38 writes the data to a MongoDB collection. I wasn't sure there was any point in also reading back the same data which was just written, but if you think it's better to have an example of .read as well as .write I'll add it back in.

rozza: Agreed its not the best of examples! That said some users will want a copy and past example of loading data.

steveren: Good point. I updated it to read into a dataframe from the MongoDB collection.

PR #17, opened 2017-02-02
More examples for reading & writing dataframe

raghothams:
Adding more examples to showcase reading & writing dataframes.
The input & output URI can be specified at the time of reading / writing dataframe instead of starting pyspark


rozza:
Thanks @raghothams

PR #19, opened 2017-03-30
Support composite _id partitionKey

ah-:
The current partitioner fails if the partitionKey is part of _id.
For example, if storing timeseries data with a schema like this:

`{"_id": {"metric": "abc", "timestamp": "2017-03-30T11:12:55"}, "value": 1.23}`

For my application I need all values for a metric in the same partition (and I can ensure there aren't too many values), so that aggregation over all values for that metric doesn't require shuffling.

To do that I want to use `_id.metric` as the partitionKey. However with the current implementation that fails, because it constructs the partition ranges as `"_id.metric": {"$lte": {"metric": "abc"}, "$gt": ...}`, resulting in an empty rdd.
Correct would be `"_id.metric": {"$lte": "abc", "$gt": ...}`.

This implements a minimal version of support for such partitionKeys.

### DIFF
--- a/src/main/scala/com/mongodb/spark/rdd/partitioner/MongoSamplePartitioner.scala
+++ b/src/main/scala/com/mongodb/spark/rdd/partitioner/MongoSamplePartitioner.scala
@@ -71,6 +71,7 @@ class MongoSamplePartitioner extends MongoPartitioner {
    * @param readConfig the [[com.mongodb.spark.config.ReadConfig]]
    * @return the partitions
    */
+  // scalastyle:off cyclomatic.complexity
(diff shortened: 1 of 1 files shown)


rozza:
Hi @ah-,

Thanks for the PR. We currently support partitioning on composite keys. 

However, there are no plans to support partitioning on parts of a composite key. The reasons for this is we can't guarantee unique ranges for the partitions. This means that we could produce overlapping partitions which would ultimately cause incorrect results.

If you know that a section of your composite key is guaranteed to plan nicely then you can always use a custom partitioner.

Ross

PR #22, opened 2017-07-18
add update to save function, if document has "_id" field

hevensun:
if the document has "_id" field, then save RDD function use update method to update the document, and the update need to has the "$set" field


rozza:
Hi @hevensun, there are no current plans to update the 1.x branch and release a new version for Spark 1.6.x

PR #23, opened 2017-07-18
add update function to save method,if the document has field "_id"

hevensun:
add update function to save method,if the document has field "_id" else use insert

### DIFF
--- a/src/main/scala/com/mongodb/spark/MongoSpark.scala
+++ b/src/main/scala/com/mongodb/spark/MongoSpark.scala
@@ -116,7 +116,7 @@ object MongoSpark {
     val mongoConnector = MongoConnector(writeConfig.asOptions)
     rdd.foreachPartition(iter => if (iter.nonEmpty) {
       mongoConnector.withCollectionDo(writeConfig, { collection: MongoCollection[Document] =>
-        iter.grouped(DefaultMaxBatchSize).foreach(batch => {
+        iter.grouped(writeConfig.maxBatchSize).foreach(batch => {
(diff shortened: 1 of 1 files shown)


hevensun:
delete white space at the end of the line, pls re ci build

hevensun:
How can I change the code to adapt to BsonDocument or Document use of D:ClassTag  in this method save[D: ClassTag](rdd: RDD[D], writeConfig: WriteConfig)? Anyone can help?

rozza:
Hi @hevensun,

Thanks for the ticket, currently the test cases are failing and this new functionality will eventually require new test cases, if it is to be merged into master.  

The issue here is an RDD is of *any* type. The difference between an RDD and a Dataset, is we know the structure of the data so thats why it is a cheap and efficient operation for Datasets as opposed to RDDs.  However, as long as there is a codec for that type then it can be serialized to MongoDB.  So the current, approach has to be able to handle that.  To normalize the types you could use the `BsonDocumentWrapper` class, along with a codec. However, the downside would be the extra encoding costs, as internally it involves converting the type into a `BsonDocument` type.

So the question is; how important is it to be able to upsert / update / replace documents from RDD's?  Please could you open a [Jira ticket](https://jira.mongodb.org/browse/SPARK) so we can explore this feature request further?

hevensun:
@rozza I open a Jira with url  https://jira.mongodb.org/browse/SPARK-130

rozza:
Closing as this relies on an unsafe cast to `BsonDocument`.  With an `RDD[D]` you cannot assume that the type `D` is a `BsonDocument`.

PR #25, opened 2018-01-17
Fix null issue

ofrapa:
See issue spark-158
Transform BsonType.NULL to null. Previously it was handled by the default case - transformed to json, so the result was the string "null".


rozza:
Thanks @ofrapa this has been pushed to master for the 2.2.2 release.

PR #36, opened 2019-04-03
[Spark][SQL]Add java spi for mongo.

beliefer:
Spark support custom data source connector and using java `spi` to register.
See at: https://github.com/databricks/spark-xml/blob/a0a8b15d940bdc91eb7d6d24e8fdc60baf33a2a5/src/main/scala/com/databricks/spark/xml/DefaultSource.scala#L38

See also https://github.com/databricks/spark-xml/blob/a0a8b15d940bdc91eb7d6d24e8fdc60baf33a2a5/src/main/resources/META-INF/services/org.apache.spark.sql.sources.DataSourceRegister

Spark original code contains some code of java `spi` as follows:
```
  def lookupDataSource(provider: String, conf: SQLConf): Class[_] = {
    val provider1 = backwardCompatibilityMap.getOrElse(provider, provider) match {
      case name if name.equalsIgnoreCase("orc") &&
          conf.getConf(SQLConf.ORC_IMPLEMENTATION) == "native" =>
        classOf[OrcFileFormat].getCanonicalName
      case name if name.equalsIgnoreCase("orc") &&
          conf.getConf(SQLConf.ORC_IMPLEMENTATION) == "hive" =>
        "org.apache.spark.sql.hive.orc.OrcFileFormat"
      case name => name
    }
    val provider2 = s"$provider1.DefaultSource"
    val loader = Utils.getContextOrSparkClassLoader
    val serviceLoader = ServiceLoader.load(classOf[DataSourceRegister], loader)

    try {
      serviceLoader.asScala.filter(_.shortName().equalsIgnoreCase(provider1)).toList match {
```

But `mongo` connector still not implement the function.
So I add this `spi` configuration.

### DIFF
--- a/src/main/resources/META-INF/services/org.apache.spark.sql.sources.DataSourceRegister
+++ b/src/main/resources/META-INF/services/org.apache.spark.sql.sources.DataSourceRegister
@@ -1,0 +1,1 @@
+com.mongodb.spark.sql.DefaultSource


jaceklaskowski:
A good catch! I'd also add a note to the docs that (once the PR gets accepts) the mongo connector would be available using `format("mongo")` (rather than the fully-qualified class name of the data source provider), e.g. https://docs.mongodb.com/spark-connector/current/scala/datasets-and-sql/

rozza:
Thanks @beliefer I'll add for the next release :+1:

beliefer:
> A good catch! I'd also add a note to the docs that (once the PR gets accepts) the mongo connector would be available using `format("mongo")` (rather than the fully-qualified class name of the data source provider), e.g. https://docs.mongodb.com/spark-connector/current/scala/datasets-and-sql/

Thanks for your supplement! I also think so.

beliefer:
> Thanks @beliefer I'll add for the next release

Thanks rozza! I am very happy to contribute to this project.

rozza:
Rebased and merged in f630feb75fd0790fa6a66eb2de1a870277ec32cf

Many thanks @beliefer :)

beliefer:
@rozza Thank you too.

PR #45, opened 2021-01-27
Added replace and update support for RDD's

rozza:
If the RDD contains the `_id` field or the shard key then
when saving values replace and upserts will be used.

Use `WriteConfig.forceInserts` to use inserts rather than upserts.

SPARK-280

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Mongo Spark Connector Changelog
 
+## 3.0.1
+  * [[SPARK-280](https://jira.mongodb.org/browse/SPARK-280)] Support replace and updates from RDD's that include an `_id`. 
+    Use `forceInsert` to keep existing behaviour.
+
 ## 3.0.0
   * [[SPARK-274](https://jira.mongodb.org/browse/SPARK-274)] Spark 3.0.0 support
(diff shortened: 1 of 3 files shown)


rozza:
https://spruce.mongodb.com/version/6011878a850e612e0c633071

stIncMale:
My understanding is that there are two `MongoSpark.save` methods, one for saving objects of type `Dataset `, and one for saving objects of type `RDD`. Previously the "if not force insert then replace or update" logic was only implemented for `Dataset` objects, and this PR implements that logic also for `RDD` objects (to avoid code duplication, `Dataset` objects are converted to `RDD` objects before saving them). Is this understanding correct?

I do not have any concerns with regard to this PR, but I am not in a state (no Scala knowledge, not familiar with mongo-spark connector) to meaningfully approve the PR.

rozza:
@stIncMale no problems - at some point we'll get you up to date with Spark! 

Datasets are backed by RDD's - so both methods get the underlying RDD and use them.  RDD's can be thought of as a distributed `Stream<T>` where type `T` can be anything. Datasets are special because they have schema metadata and only contain fixed Spark types.

jyemin (approved):
LGTM

PR #47, opened 2021-04-20
Spark 3.1.1 support

rozza:
SPARK-293

https://spruce.mongodb.com/version/607ecc385623436443a9f4c6

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Mongo Spark Connector Changelog
 
+## 3.1.0
+  * [[SPARK-293](https://jira.mongodb.org/browse/SPARK-293)] Spark 3.1.1 support
+
 ## 3.0.1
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Mongo Spark Connector Changelog
 
+## 3.1.0
```
jaceklaskowski, line 3: nit: 3.1.1? Why not to match Spark version?

jaceklaskowski (commented):
I must admit I'm surprised how few changes were proposed to support Spark 3.1.1. Nothing bad but expected `Table` and such. Anyway...

CHANGELOG.md
rozza: Because this is the first release of the 3.1 series and there have been no patches to the connector.

rozza:
@jaceklaskowski my tests have shown the 3.0.1 spark connector works against 3.1.1 Spark. Which is how semantic versioning should work but unfortunately hasn't always been the case with previous minor spark releases.

PR #49, opened 2021-07-29
Initial build

rozza:
The initial spark connector build configuration.

Based off the Kafka build. 

Note changed package from the older spark connector: from `com.mongodb.spark` to `com.mongodb.spark.connector` so that both could be run at the same time.


rozza:
Closed see #50

PR #50, opened 2021-07-29
The initial spark connector build configuration.

rozza:
Based off the Kafka build. 

Note: changed package from the older spark connector: from `com.mongodb.spark` to `com.mongodb.spark.connector` so that both could be run at the same time.


build.gradle.kts
```diff
@@ -0,0 +1,348 @@
...
+plugins {
+    idea
+    `java-library`
+    `maven-publish`
+    signing
+    checkstyle
+    id("com.github.gmazzo.buildconfig") version "3.0.2"
+    id("com.github.spotbugs") version "4.7.2"
+    id("com.diffplug.spotless") version "5.14.2"
+    id("com.github.johnrengelman.shadow") version "6.1.0"
```
jyemin, line 36: Are we going to be shadowing anything?

build.gradle.kts
```diff
@@ -0,0 +1,348 @@
...
+    sourceCompatibility = JavaVersion.VERSION_1_8
+    targetCompatibility = JavaVersion.VERSION_1_8
+}
+
+repositories {
+    mavenCentral()
+    maven("https://jitpack.io")
+}
+
+extra.apply {
```
jyemin, line 53: What's `extra`?

build.gradle.kts
```diff
@@ -0,0 +1,348 @@
...
+    set("mongodbDriverVersion", "[4.3,4.3.99)")
```
jyemin, line 54: Is the trailing `.0` implied?

build.gradle.kts
```diff
@@ -0,0 +1,348 @@
...
+    set("sparkVersion", "3.1.2")
+
+    // Testing dependencies
+    set("junitJupiterVersion", "5.7.2")
+    set("junitPlatformVersion", "1.7.2")
+
+    // Integration test dependencies
+}
+
+val mongoDependencies: Configuration by configurations.creating
```
jyemin, line 64: What's this for?

build.gradle.kts
```diff
@@ -0,0 +1,348 @@
...
+
+dependencies {
+    implementation("org.apache.spark:spark-parent_2.12:${project.extra["sparkVersion"]}")
+    implementation("org.mongodb:mongodb-driver-sync:${project.extra["mongodbDriverVersion"]}")
+
+    mongoDependencies("org.mongodb:mongodb-driver-sync:${project.extra["mongodbDriverVersion"]}")
```
jyemin, line 70: Why do we have to declare the driver dependency twice?

build.gradle.kts
```diff
@@ -0,0 +1,348 @@
...
+}
+
+tasks.withType<JavaCompile> {
+    options.encoding = "UTF-8"
+}
+
+/*
+ * Generated build config
+ */
+val gitVersion: String by lazy {
```
jyemin, line 86: Our old friend

build.gradle.kts
```diff
@@ -0,0 +1,348 @@
...
+tasks.named("compileJava") {
+    dependsOn(":spotlessApply")
+}
+
+/*
+ * ShadowJar
+ */
+tasks.register<ShadowJar>("allJar") {
+    archiveClassifier.set("all")
+    from(mongoDependencies, sourceSets.main.get().output)
```
jyemin, line 220: Why do we need to shadow the driver?

jyemin (commented):
Looks good.  Just a bunch of questions, really

build.gradle.kts
rozza: Yes it is implied. Its > 4.3.0 && < 4.3.99 

We don't do < 4.4 as that will pick up 4.4.0-beta1

build.gradle.kts
rozza: This will be used to build the all jar which users will use with Spark.

Deploying to spark (and Kafka) doesn't use maven do its just add the jar file to the plugins location. So we need to create an uber jar to do that and make it easy.

build.gradle.kts
rozza: This is for the uber jar that we'll also publish.

build.gradle.kts
rozza: The Shadow Jar plugin is used to build the uber all jar.

build.gradle.kts
rozza: Its a map where we declare the dependencies. Its neat because it keeps all the dependencies in one place and ensures that the same version is used for all the spark dependencies

rozza: An alternative would be to do something like: 
https://docs.gradle.org/current/userguide/platforms.html#sub:conventional-dependencies-toml

build.gradle.kts
rozza: No plans to, shadow will be used to build the uber jar.

jyemin (approved):
Thanks for all the explanations.  

LGTM

PR #54, opened 2021-09-17
Added Row to BsonDocument conversion

rozza:
SPARK-310

### DIFF
--- a/src/main/java/com/mongodb/spark/sql/connector/schema/ConverterHelper.java
+++ b/src/main/java/com/mongodb/spark/sql/connector/schema/ConverterHelper.java
@@ -29,9 +29,9 @@ import org.bson.codecs.Codec;
 import org.bson.json.JsonMode;
 import org.bson.json.JsonWriterSettings;
 
-interface ConverterHelper {
-  Codec<BsonValue> BSON_VALUE_CODEC = new BsonValueCodec();
-  JsonWriterSettings DEFAULT_JSON_WRITER_SETTINGS =
+final class ConverterHelper {
+  static final Codec<BsonValue> BSON_VALUE_CODEC = new BsonValueCodec();
+  static final JsonWriterSettings DEFAULT_JSON_WRITER_SETTINGS =
       JsonWriterSettings.builder()
           .outputMode(JsonMode.RELAXED)
           .binaryConverter(
@@ -47,11 +47,5 @@ interface ConverterHelper {
           .symbolConverter((value, writer) -> writer.writeString(value))
           .build();
 
-  static Codec<BsonValue> getBsonValueCodec() {
-    return BSON_VALUE_CODEC;
-  }
-
-  static JsonWriterSettings getDefaultJsonWriterSettings() {
-    return DEFAULT_JSON_WRITER_SETTINGS;
-  }
+  private ConverterHelper() {}
 }
--- a/src/main/java/com/mongodb/spark/sql/connector/schema/ConverterHelper.java
+++ b/src/main/java/com/mongodb/spark/sql/connector/schema/ConverterHelper.java
@@ -1,0 +1,57 @@
+/*
+ * Copyright 2008-present MongoDB, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package com.mongodb.spark.sql.connector.schema;
+
+import java.time.Instant;
+import java.time.ZoneOffset;
+import java.time.ZonedDateTime;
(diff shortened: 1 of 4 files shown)


rozza:
Its rewritten code from the original scala version.

src/main/java/com/mongodb/spark/sql/connector/schema/ConverterHelper.java
```diff
@@ -0,0 +1,57 @@
...
+              (value, writer) -> {
+                ZonedDateTime zonedDateTime = Instant.ofEpochMilli(value).atZone(ZoneOffset.UTC);
+                writer.writeString(DateTimeFormatter.ISO_DATE_TIME.format(zonedDateTime));
+              })
+          .decimal128Converter((value, writer) -> writer.writeString(value.toString()))
+          .objectIdConverter((value, writer) -> writer.writeString(value.toHexString()))
+          .symbolConverter((value, writer) -> writer.writeString(value))
+          .build();
+
+  static Codec<BsonValue> getBsonValueCodec() {
```
stIncMale, line 50: We should convert `ConverterHelper` to a `final` class with a private constructor for the following reasons:

1. The ability to define `static` methods in interfaces was added to Java because it [_"makes it easier for you to organize helper methods in your libraries; you can keep static methods specific to an interface in the same interface rather than in a separate class"_](https://docs.oracle.com/javase/tutorial/java/IandI/defaultmethods.html#static). However, this is not applicable to the interface `ConverterHelper`: the interface is meant to be solely a collection of `static` methods and is not meant to define a type.
2. The purpose of `static` methods `getBsonValueCodec`/`getDefaultJsonWriterSettings` appear to be to provide an identity-agnostic way of getting a BSON value codec / default JSON writer settings. However, the fact that the `static` fields `BSON_VALUE_CODEC`/`DEFAULT_JSON_WRITER_SETTINGS` are also `public`, diminishes the benefit of these methods, because a user can still access the BSON value codec / default JSON writer settings in a way that then allows to rely on their identity.

Generally speaking, an interface may be used only if it is meant to define a type, i.e., to be implemented.

src/test/java/com/mongodb/spark/sql/connector/schema/RowToBsonDocumentConverterTest.java
```diff
@@ -0,0 +1,211 @@
...
+
+  private <K, V> scala.collection.Map<K, V> toMap(final K key, final V value) {
+    Map<K, V> map = new HashMap<>();
+    map.put(key, value);
+    return JavaConverters.mapAsScalaMap(map);
+  }
+
+  private scala.collection.Map<?, ?> toMap(final Map<?, ?> map) {
+    return JavaConverters.mapAsScalaMap(map);
+  }
```
stIncMale, line 210: It seems, this method is not used and can be removed.

src/main/java/com/mongodb/spark/sql/connector/schema/RowToBsonDocumentConverter.java
```diff
@@ -0,0 +1,131 @@
...
+import org.bson.BsonDouble;
+import org.bson.BsonInt32;
+import org.bson.BsonInt64;
+import org.bson.BsonString;
+import org.bson.BsonValue;
+import org.bson.types.Decimal128;
+
+import com.mongodb.spark.sql.connector.exceptions.DataException;
+
+interface RowToBsonDocumentConverter {
```
stIncMale, line 53: I suggest changing this interface to a class for mostly the same reasons as in this [comment](https://github.com/mongodb/mongo-spark/pull/54/files#r716296848).

stIncMale (changes requested):

src/main/java/com/mongodb/spark/sql/connector/schema/RowToBsonDocumentConverter.java
rozza: Done

src/main/java/com/mongodb/spark/sql/connector/schema/ConverterHelper.java
rozza: I like it better as you suggested - thanks @stIncMale

src/main/java/com/mongodb/spark/sql/connector/schema/RowToBsonDocumentConverter.java
stIncMale: I do not see related changes in the second commit, maybe they were not pushed.

But the above observation is good:), because I also noticed that `BsonDocumentToRowConverter` is an instantiable class, while `RowToBsonDocumentConverter` is a collection of static methods. I understand that `RowToBsonDocumentConverter` does not need to be instantiable, but the inconsistency in what a converter is (an object or a `static` method) is not good. I would like to change the [previous proposal](https://github.com/mongodb/mongo-spark/pull/54#discussion_r716311756) and propose to consider changing this interface not to a class with `static` methods, but to an instantiable class for consistency with `BsonDocumentToRowConverter`.

jyemin:
@rozza I'm going to leave this to @stIncMale so feel free to merge after he approves.

src/main/java/com/mongodb/spark/sql/connector/schema/RowToBsonDocumentConverter.java
rozza: A happy accident! I've made both converters final classes. Thanks @stIncMale

stIncMale (approved):

jyemin:
@rozza one LGTM is sufficient for this one.

PR #55, opened 2021-09-23
Added MongoClientCache

rozza:
Built upon SPARK-302.

MongoClientCache is the main code to review.
MongoConfig - is the start of the configuration for the connector.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -57,6 +57,7 @@ extra.apply {
     // Testing dependencies
     set("junitJupiterVersion", "5.7.2")
     set("junitPlatformVersion", "1.7.2")
+    set("mockitoVersion", "3.12.4")
 
     // Integration test dependencies
 }
@@ -76,6 +77,7 @@ dependencies {
     // Unit Tests
     testImplementation(platform("org.junit:junit-bom:5.7.2"))
     testImplementation("org.junit.jupiter:junit-jupiter")
+    testImplementation("org.mockito:mockito-junit-jupiter:${project.extra["mockitoVersion"]}")
 
     // Integration Tests
 }
--- a/src/integrationTest/java/com/mongodb/spark/sql/connector/MongoCatalogTest.java
+++ b/src/integrationTest/java/com/mongodb/spark/sql/connector/MongoCatalogTest.java
@@ -55,6 +55,8 @@ import com.mongodb.client.MongoDatabase;
 import com.mongodb.client.model.Aggregates;
 import com.mongodb.client.model.Filters;
 
+import com.mongodb.spark.sql.connector.config.MongoConfig;
+import com.mongodb.spark.sql.connector.connection.MongoConnectionProvider;
 import com.mongodb.spark.sql.connector.mongodb.MongoSparkConnectorTestCase;
 
 public class MongoCatalogTest extends MongoSparkConnectorTestCase {
@@ -306,7 +308,8 @@ public class MongoCatalogTest extends MongoSparkConnectorTestCase {
         new MongoTable(
             new MongoNamespace(TEST_DATABASE_NAME, TEST_COLLECTION_NAME),
             null,
-            getConnectionProviderOptions()),
+            new MongoConnectionProvider(
+                MongoConfig.createInputConfig(getConnectionProviderOptions()))),
         MONGO_CATALOG.loadTable(TEST_IDENTIFIER));
   }
 
@@ -358,7 +361,10 @@ public class MongoCatalogTest extends MongoSparkConnectorTestCase {
 
     assertEquals(
         new MongoTable(
-            new MongoNamespace(TEST_DATABASE_NAME, TEST_COLLECTION_NAME), schema, emptyMap()),
+            new MongoNamespace(TEST_DATABASE_NAME, TEST_COLLECTION_NAME),
+            schema,
+            new MongoConnectionProvider(
+                MongoConfig.createInputConfig(getConnectionProviderOptions()))),
         MONGO_CATALOG.createTable(TEST_IDENTIFIER, schema, new Transform[0], emptyMap()));
   }
 
--- a/src/main/java/com/mongodb/spark/sql/connector/config/ClassHelper.java
+++ b/src/main/java/com/mongodb/spark/sql/connector/config/ClassHelper.java
@@ -1,0 +1,97 @@
+/*
+ * Copyright 2008-present MongoDB, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package com.mongodb.spark.sql.connector.config;
+
+import static java.lang.String.format;
+
+import java.util.List;
+
+import org.jetbrains.annotations.ApiStatus;
+
+import com.mongodb.spark.sql.connector.exceptions.ConfigException;
+
+@ApiStatus.Internal
+interface ClassHelper {
+
+  @SuppressWarnings("unchecked")
+  static <T> T createInstance(
+      final String configKey, final String className, final Class<T> clazz) {
+    return createInstance(
+        configKey,
+        className,
+        clazz,
+        () -> (T) Class.forName(className).getConstructor().newInstance());
+  }
+
+  @SuppressWarnings("unchecked")
+  static <T> T createInstance(
+      final String configKey,
+      final String className,
+      final Class<T> clazz,
+      final List<Class<?>> constructorArgs,
+      final List<Object> initArgs) {
+    return createInstance(
+        configKey,
+        className,
+        clazz,
+        () ->
+            (T)
+                Class.forName(className)
+                    .getConstructor(constructorArgs.toArray(new Class<?>[0]))
+                    .newInstance(initArgs.toArray(new Object[0])));
+  }
+
+  static <T> T createInstance(
+      final String configKey,
+      final String className,
+      final Class<T> clazz,
+      final ClassCreator<T> cc) {
+    try {
+      return cc.init();
+    } catch (ClassCastException e) {
+      throw new ConfigException(
+          configKey,
+          className,
+          format("Contract violation class doesn't implement: '%s'", clazz.getSimpleName()));
+    } catch (ClassNotFoundException e) {
+      throw new ConfigException(
+          configKey, className, format("Class not found: %s", e.getMessage()));
+    } catch (NoSuchMethodException e) {
+      throw new ConfigException(
+          configKey,
+          className,
+          format(
+              "Class could not be initialized, no public default constructor: %s", e.getMessage()));
+    } catch (Exception e) {
+      if (e.getCause() instanceof ConfigException) {
+        throw (ConfigException) e.getCause();
+      }
+      throw new ConfigException(configKey, className, e.getMessage());
+    }
+  }
+
+  static <T extends Configurable> T configureInstance(final T instance, final MongoConfig config) {
+    instance.configure(config);
+    return instance;
+  }
+
+  @FunctionalInterface
+  interface ClassCreator<T> {
+    T init() throws Exception;
+  }
+}
--- a/src/main/java/com/mongodb/spark/sql/connector/config/Configurable.java
+++ b/src/main/java/com/mongodb/spark/sql/connector/config/Configurable.java
@@ -1,0 +1,29 @@
+/*
+ * Copyright 2008-present MongoDB, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package com.mongodb.spark.sql.connector.config;
+
+/** Represents a configurable class */
+public interface Configurable {
+
+  /**
+   * Configure this class with the {@link MongoConfig}.
+   *
+   * @param config the current configuration
+   */
+  void configure(MongoConfig config);
+}
--- a/src/main/java/com/mongodb/spark/sql/connector/config/MongoConfig.java
+++ b/src/main/java/com/mongodb/spark/sql/connector/config/MongoConfig.java
@@ -1,0 +1,292 @@
+/*
+ * Copyright 2008-present MongoDB, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package com.mongodb.spark.sql.connector.config;
+
+import java.io.Serializable;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+
+import org.apache.spark.sql.util.CaseInsensitiveStringMap;
+import org.jetbrains.annotations.ApiStatus;
+import org.jetbrains.annotations.TestOnly;
+
+import com.mongodb.ConnectionString;
+
+import com.mongodb.spark.sql.connector.connection.DefaultMongoClientFactory;
+import com.mongodb.spark.sql.connector.connection.MongoClientFactory;
+
+/**
+ * The MongoConfig class
+ *
+ * <p>Provides MongoDB specific configuration. Wraps the {@link CaseInsensitiveStringMap} options
+ * provided by Spark.
+ */
+public final class MongoConfig implements Serializable {
+
+  static final long serialVersionUID = 1L;
+
+  /**
+   * The prefix for all general Spark MongoDB configurations.
+   *
+   * <p>For example the {@link MongoConfig#MONGO_CONNECTION_STRING_CONFIG} should be defined as:
+   * "{@code spark.mongodb.connection.uri}".
+   */
+  public static final String MONGO_PREFIX = "spark.mongodb.";
+
+  /**
+   * The prefix for specific input based configurations.
+   *
+   * <p>Overrides any configurations that just use the {@link MongoConfig#MONGO_PREFIX}. For example
+   * to override the {@link MongoConfig#MONGO_CONNECTION_STRING_CONFIG} just for inputting data into
+   * MongoDB: "{@code spark.mongodb.input.connection.uri}".
+   */
+  public static final String MONGO_INPUT_PREFIX = "spark.mongodb.input.";
+
+  /**
+   * The prefix for specific output based configurations.
+   *
+   * <p>Overrides any configurations that just use the {@link MongoConfig#MONGO_PREFIX}. For example
+   * to override the {@link MongoConfig#MONGO_CONNECTION_STRING_CONFIG} just for outputting data
+   * from MongoDB: "{@code spark.mongodb.output.connection.uri}".
+   */
+  public static final String MONGO_OUTPUT_PREFIX = "spark.mongodb.output.";
+
+  /**
+   * The MongoClientFactory configuration
+   *
+   * <p>The default implementation uses the {@link MongoConfig#MONGO_CONNECTION_STRING_CONFIG} as
+   * the connection string.
+   *
+   * <p>Custom implementations are allowed and must implement the {@link
+   * com.mongodb.spark.sql.connector.connection.MongoClientFactory} interface.
+   */
+  public static final String MONGO_CLIENT_FACTORY_CONFIG = "mongoClientFactory";
+  /**
+   * The default MongoClientFactory configuration
+   *
+   * <p>Requires the {@link MongoConfig#MONGO_CONNECTION_STRING_CONFIG} for configuring the
+   * resulting {@link com.mongodb.client.MongoClient}
+   */
+  public static final String MONGO_CLIENT_FACTORY_DEFAULT =
+      DefaultMongoClientFactory.class.getName();
+
+  /** The connection string configuration */
+  public static final String MONGO_CONNECTION_STRING_CONFIG = "connection.uri";
+  /** The default connection string configuration */
+  public static final String MONGO_CONNECTION_STRING_DEFAULT = "mongodb://localhost:27017/";
+
+  /** The current usage mode for the configuration. */
+  private enum UsageMode {
+    INPUT,
+    OUTPUT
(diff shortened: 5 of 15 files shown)


build.gradle.kts
```diff
@@ -76,6 +77,7 @@ dependencies {
     // Unit Tests
     testImplementation(platform("org.junit:junit-bom:5.7.2"))
     testImplementation("org.junit.jupiter:junit-jupiter")
+    testImplementation("org.mockito:mockito-junit-jupiter:${project.extra["mockitoVersion"]}")
```
rozza, line 80: Added for unit testing the cache.

src/integrationTest/java/com/mongodb/spark/sql/connector/MongoCatalogTest.java
```diff
@@ -306,7 +308,8 @@ void loadTableTest() throws NoSuchTableException {
         new MongoTable(
             new MongoNamespace(TEST_DATABASE_NAME, TEST_COLLECTION_NAME),
             null,
-            getConnectionProviderOptions()),
+            new MongoConnectionProvider(
```
rozza, line 311: The `MongoConnectionProvider` now takes the `MongoConfig` rather than the raw options.

src/main/java/com/mongodb/spark/sql/connector/config/ClassHelper.java
```diff
@@ -0,0 +1,97 @@
...
+import static java.lang.String.format;
+
+import java.util.List;
+
+import org.jetbrains.annotations.ApiStatus;
+
+import com.mongodb.spark.sql.connector.exceptions.ConfigException;
+
+@ApiStatus.Internal
+interface ClassHelper {
```
rozza, line 29: Copied from the Kafka connector

src/main/java/com/mongodb/spark/sql/connector/config/Configurable.java
```diff
@@ -0,0 +1,29 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package com.mongodb.spark.sql.connector.config;
+
+/** Represents a configurable class */
+public interface Configurable {
```
rozza, line 21: Copied from the Kafka connector

src/main/java/com/mongodb/spark/sql/connector/config/MongoConfig.java
```diff
@@ -0,0 +1,292 @@
...
+                defaults.add(k);
+              }
+            });
+
+    defaults.forEach(
+        k -> usageSpecificSettings.put(k.substring(MONGO_PREFIX.length() - 1), originals.get(k)));
+    overrides.forEach(
+        k -> usageSpecificSettings.put(k.substring(overridePrefix.length() - 1), originals.get(k)));
+
+    this.values = new CaseInsensitiveStringMap(usageSpecificSettings);
```
rozza, line 154: Override any default configuration using usage mode specific configuration.

src/main/java/com/mongodb/spark/sql/connector/connection/DefaultMongoClientFactory.java
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  /** @return create a new instance of a {@code MongoClient}. */
+  @Override
+  public MongoClient create() {
+    return MongoClients.create(config.getConnectionString());
+  }
+
+  @Override
+  public boolean equals(final Object o) {
```
rozza, line 58: equals is required for the MongoClientCache to work

src/main/java/com/mongodb/spark/sql/connector/connection/MongoClientCache.java
```diff
@@ -0,0 +1,179 @@
...
+
+/**
+ * A simple cache that handles the acquisition and release of {@code MongoClients}.
+ *
+ * <p>Handles the creation and closing of {@code MongoClients}. When a {@link MongoClient} is no
+ * longer used there is a grace period allowing the {@code MongoClient} to be reused. As Spark
+ * Executors can operate on multiple tasks this reduces the cost creating a MongoClient and
+ * connections.
+ */
+final class MongoClientCache {
```
rozza, line 43: This code is much more simple than the previous connectors cache. It does use synchronization locking on acquisition, release and cleanup of clients. This shouldn't be impediment as they are cheap operations.

src/main/java/com/mongodb/spark/sql/connector/connection/MongoConnectionProvider.java
```diff
@@ -82,5 +144,56 @@ public void doWithCollection(
         mongoDatabase ->
             function.apply(mongoDatabase.getCollection(collectionName, BsonDocument.class)));
   }
-  // CHECKSTYLE.ON:JavadocMethod
+
+  /** @return memoize's and returns the {@link MongoClientCache} */
+  private static synchronized MongoClientCache getMongoClientCache() {
```
rozza, line 149: Creates the singleton MongoClientCache if it doesn't already exist.

Uses the same system property as the existing spark connector for configuring the keep alive period

stIncMale: A couple of months ago a similar lazy initialization approach in the BSON library was reported to cause performance problems to a client, and the client [proposed](https://github.com/mongodb/mongo-java-driver/pull/759) to use the [DCL](https://twitter.com/shipilev/status/1229730431223975936?lang=en) (the version mentioned in the linked tweet is rather a "double-locked checking" 😄). Do we want to consider using the DCL here to avoid a similar situation again?

Alternatively, we can remove the method `getMongoClientCache()` and utilize the [Java class lazy initialization rules](https://docs.oracle.com/javase/specs/jls/se8/html/jls-12.html#jls-12.4.1):
```java
private static final class Lazy {
    static MongoClientCache mongoClientCache;

    static {
        int keepAliveMS = 5000;
        try {
            keepAliveMS =
                    Integer.parseInt(System.getProperty(SYSTEM_MONGO_CACHE_KEEP_ALIVE_MS_PROPERTY, "5000"));
        } catch (NumberFormatException e) {
            // ignore and use default
        }
        mongoClientCache = new MongoClientCache(keepAliveMS);
        Runtime.getRuntime().addShutdownHook(new ShutdownHook(mongoClientCache));
    }
}
```
With the above class, we can replace all invocations of `getMongoClientCache()` with `Lazy.mongoClientCache`, and it will be guaranteed that a `MongoClientCache` is created only when the execution actually tries to read from the `mongoClientCache` field.

stIncMale (changes requested):
Reviewed everything but `MongoClientCache`, which I will review tomorrow. Submitting the review comments now as there is a number of them, and I am not sure if GitHub may lose them if they are not submitted.

rozza (commented):
Thanks @stIncMale for the initial review.

For your edification, like Kafka most / all this code won't be used directly by a user. Rather they will use the Spark API which will then call things from the MongoCatalog / MongoTable.
Some specialized usecases willl still exist, for users partitioning their own collections when reading data etc. and those users will extend or implement  the relative interfaces.

I have removed all serialization extensions for the config / connections - thank you for pointing that out (especially that CaseInsensitiveMap isn't serializiable). Certain parts will need to be serializable in future PRs as units of work will be serialized from the Executor to the Workers.

src/main/java/com/mongodb/spark/sql/connector/connection/MongoConnectionProvider.java
```diff
@@ -41,12 +40,9 @@
  * MongoClientCache}.
  */
 @ApiStatus.Internal
-public final class MongoConnectionProvider implements Serializable {
+@ThreadSafe
+public final class MongoConnectionProvider {
```
stIncMale, line 44: `MongoConnectionProvider` must be thread-safe, but `MongoConfig` is not thread-safe, and `MongoConnectionProvider.withClient` calls `MongoConfig.getMongoClientFactory`, which is a contradiction.

P.S.
You may say that I proposed to mark `MongoConfig` as not thread-safe in https://github.com/mongodb/mongo-spark/pull/55#discussion_r720911762, but I did that based on its code (as I mentioned) and not its usage. At that time it was unspecified whether `MongoConnectionProvider` must be thread-safe or not, so I could not judge based on the `MongoConfig` usage.

stIncMale (changes requested):

src/main/java/com/mongodb/spark/sql/connector/config/MongoConfig.java
```diff
@@ -0,0 +1,292 @@
...
+   *
+   * @param key the key whose associated value is to be returned
+   * @param defaultValue the default mapping for the config
+   * @return the value to which the specified key is mapped, or {@code defaultValue} if this config
+   *     contains no mapping for the key
+   * @throws ClassCastException if the key is of an inappropriate type for this map
+   * @throws NullPointerException if the specified key is null and this map does not permit null
+   *     keys
+   */
+  public String getOrDefault(final Object key, final String defaultValue) {
```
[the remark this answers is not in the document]
stIncMale: It appears that we still need to specify that `defaultValue` is allowed to be `null`.

src/main/java/com/mongodb/spark/sql/connector/connection/MongoClientCache.java
stIncMale: > This shouldn't be an impediment as they are cheap operations.

The algorithmic complexity of the method `checkClientCache` is O(numberOfCachedItems), which while may be considered cheap in some scenarios, it does not seem like something one may expect to be cheap in the absolute majority of use cases. And other methods are blocked by this linear-time method.

I understand that the decision to make all methods synchronized makes the implementation simpler. But this does not explain why we use `ArrayList` and linear-time search instead of using `HashMap` and constant-time search when accessing the cache. This decision makes the time complexity of all methods O(numberOfCachedItems), which is worrisome given that they also block each other.

Given that mongo-spark is not meant to be used as a library, but as a plugin (or something like that), I expect we are free to use 3rd party dependencies. If this is the case, can we consider using https://github.com/ben-manes/caffeine to get a cache? It's what [Guava encourages to use](https://guava.dev/releases/snapshot-jre/api/docs/com/google/common/cache/CacheBuilder.html) instead of Guava's cache.

src/main/java/com/mongodb/spark/sql/connector/connection/MongoClientCache.java
```diff
@@ -0,0 +1,179 @@
...
+    }
+
+    private MongoClient acquire() {
+      referenceCount += 1;
+      releaseTime = Long.MAX_VALUE;
+      return mongoClient;
+    }
+
+    private void release() {
+      referenceCount -= 1;
```
stIncMale, line 161: Since nothing stops the code that uses `MongoClientCache` from releasing more than once per acquire, let's have an assertion here that ensures `referenceCount` does not go below 0.

stIncMale (changes requested):

src/main/java/com/mongodb/spark/sql/connector/connection/MongoClientFactory.java
```diff
@@ -19,15 +19,29 @@
...
+ * <p>A factory should produce {@code MongoClients} with the same configuration each time {@link
+ * MongoClientFactory#create()} is called.
+ *
+ * <p>The creation of a {@link MongoClient} is expensive (creating the connection pool and
+ * authenticating connections), so {@code MongoClients} are automatically cached, using the {@code
+ * MongoClientFactory} as the key.
+ *
+ * <p>Implementations of the factory can make use of the automatic caching by ensuring the {@code
+ * MongoClientFactory.equals()} returns true if the {@code MongoClientFactory} instances configured
+ * with the same {@link com.mongodb.spark.sql.connector.config.MongoConfig} and if the
```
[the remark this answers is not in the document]
rozza: Both are true. The DefaultMongoClientFactory will produce the same MongoClient per same MongoConfig. It will also produce the same client for any MongoConfig that has the same connection string.

Alternative implementations may choose to match the whole config or parts depending on their usecase.

I will add a clarification.

src/main/java/com/mongodb/spark/sql/connector/connection/MongoConnectionProvider.java
rozza: Sorry MongoConfig is ThreadSafe.

src/main/java/com/mongodb/spark/sql/connector/connection/MongoClientCache.java
```diff
@@ -0,0 +1,179 @@
...
+    private final MongoClient mongoClient;
+    private long releaseTime;
+    private int referenceCount;
+
+    private MongoClientCacheItem(
+        final MongoClientFactory mongoClientFactory, final long keepAliveNanos) {
+      this.mongoClientFactory = mongoClientFactory;
+      this.mongoClient = mongoClientFactory.create();
+      this.keepAliveNanos = keepAliveNanos;
+      this.releaseTime = Long.MAX_VALUE;
```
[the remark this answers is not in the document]
rozza: Thanks updated.

src/main/java/com/mongodb/spark/sql/connector/connection/MongoClientCache.java
```diff
@@ -0,0 +1,179 @@
...
+    cacheList.stream()
+        .filter(c -> Objects.equals(c.mongoClient, mongoClient))
+        .findFirst()
+        .orElseThrow(() -> new IllegalStateException("MongoClient not in the cache."))
+        .release();
+  }
+
+  void shutdown() {
+    if (isAvailable) {
+      scheduler.shutdown();
```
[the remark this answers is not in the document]
rozza: Updated

src/main/java/com/mongodb/spark/sql/connector/connection/MongoClientCache.java
```diff
@@ -0,0 +1,179 @@
...
+      cacheList.forEach(
+          item -> {
+            try {
+              item.mongoClient.close();
+            } catch (Exception e) {
+              LOGGER.info("Error when shutting down client: {}", e.getMessage());
+            }
+          });
+      cacheList.clear();
```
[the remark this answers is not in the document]
rozza: Shutdown is only called when the runtime shutdown hook is called and is used to free all resources. I've synchronized the method.

src/main/java/com/mongodb/spark/sql/connector/connection/MongoClientCache.java
rozza: Done.

src/main/java/com/mongodb/spark/sql/connector/connection/MongoClientCache.java
rozza: I think taking an extra dependency is complex for this as it really is in the library arena. We don't have control over the dependencies used by Spark or the users programs passed into Spark.

o(n) shouldn't be an issue as the cache is expected to be small (less than 100, in reality less than 10). The size of the cache is limited to each worker for the lifetime of a task. Its job is to reduce the cost of creating MongoClients when mulitple tasks are sent to the same Spark worker.  

It would be quite unusual for multiple differing Mongo-Spark configs to be required on the same Spark worker, at the same time, competing for the same resources. The keep alives are short enough to reduce the size of queue to make this effective enough.  

However, it can be revisited in the future should feedback or usecases change.

src/main/java/com/mongodb/spark/sql/connector/connection/MongoClientCache.java
```diff
@@ -0,0 +1,179 @@
...
+      if (referenceCount == 0) {
+        releaseTime = System.nanoTime() + keepAliveNanos;
+      }
+    }
+
+    private boolean shouldBeRemoved(final long currentTime) {
+      if (referenceCount == 0 && releaseTime < currentTime) {
+        try {
+          mongoClient.close();
+        } catch (Exception e) {
```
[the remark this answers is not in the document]
rozza: Updated.

src/main/java/com/mongodb/spark/sql/connector/connection/MongoClientCache.java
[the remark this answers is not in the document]
rozza: I'm still confused as to why releaseTime being set to MAX_VALUE is not correct.  System.nanoTime cannot be bigger that MAX_VALUE - so its used as a shortcut meaning never release.

I adapted your version, the difference is the keep alive period starts when the MongoClient is fully released. This provides the window for another task to reuse the MongoClient (reducing the costs of creating one).

rozza (commented):
PR updated @stIncMale - thanks again for the feedback :)

src/main/java/com/mongodb/spark/sql/connector/connection/MongoClientCache.java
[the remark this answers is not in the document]
stIncMale: > I adapted your version

I cannot find new commits, probably the latest changes were not pushed.

> System.nanoTime cannot be bigger that MAX_VALUE

This is false, and I think once you see why, the rest of "confused as to why releaseTime being set to MAX_VALUE is not correct" should be clear.

The minimum value of `System.nanoTime` is the [origin](https://docs.oracle.com/javase/8/docs/api/java/lang/System.html#nanoTime--), which is unknown an is determined by a JVM process. The maximum value of `System.nanoTime` is origin + 2<sup>63</sup> (may overflow, and that is fine). The total order ≤<sub>nano</sub> is defined via the total order ≤<sub>long</sub> only for values within the interval [origin; origin + 2<sup>63</sup>]: x ≤<sub>nano</sub> y ⇔ x - y ≤<sub>long</sub> 0. Note that the subtraction here is a subtraction in `long`, i.e., it can underflow/overflow.

Now, for example, if the origin is `Long.MAX_VALUE` - 1, then ≤<sub>nano</sub> is defined for `Long.MAX_VALUE` and `Long.MIN_VALUE`, and we can see that `Long.MAX_VALUE` ≤<sub>nano</sub> `Long.MIN_VALUE` because `Long.MAX_VALUE` - `Long.MIN_VALUE` = -1 ≤<sub>long</sub> 0.

In fact, only one value, the origin, is smaller than `Long.MAX_VALUE` in the above example. All the other `System.nanoTime` values, except for the `Long.MAX_VALUE` itself, are bigger than `Long.MAX_VALUE`.

P.S. In [later versions](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/System.html#nanoTime()) the documentation of the `System.nanoTime` methods was changed from trying to specify the total order to just warning against the pattern `System.nanoTime() >= startTime + timeoutNanos`, which is currently used in the `MongoClientCacheItem` in addition to relying on the assumption that `Long.MAX_VALUE` is the largest `System.nanoTime` value.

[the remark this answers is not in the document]
rozza: @stIncMale - apologies for not pushing the changes - I blame being poorly!

Also thank you for the explanation :+1:

src/main/java/com/mongodb/spark/sql/connector/connection/MongoClientFactory.java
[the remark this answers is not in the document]
stIncMale: > The DefaultMongoClientFactory will produce the same MongoClient per same MongoConfig.
> It will also produce the same client for any MongoConfig that has the same connection string.

It produces equivalent `MongoClient` objects for equal (not necessary the same) `MongoConfig`s / connection strings. These `MongoClient` objects are neither distinct nor not unequal (`MongoClientImpl` does not override `Object.equals`).

The above fact does not require any changes in the clarified `MongoClientFactory` documentation, but please keep in mind that the terms same/distinct and equal/unequal have very different meaning explained in [JLS 15.21.3](https://docs.oracle.com/javase/specs/jls/se17/html/jls-15.html#jls-15.21.3): _"While `==` may be used to compare references of type `String`, such an equality test determines whether or not the two operands refer to the **same** `String` object. The result is `false` if the operands are **distinct** `String` objects, even if they contain the same sequence of characters. The contents of two strings `s` and `t` can be tested for **equality** by the method invocation `s.equals(t)`."_

src/main/java/com/mongodb/spark/sql/connector/connection/MongoClientCache.java
```diff
@@ -139,36 +138,37 @@ private void assertIsAvailable() {
...
       referenceCount -= 1;
-      if (referenceCount == 0) {
-        releaseTime = System.nanoTime() + keepAliveNanos;
-      }
     }
 
-    private boolean shouldBeRemoved(final long currentTime) {
-      if (referenceCount == 0 && releaseTime < currentTime) {
+    private boolean shouldBeRemoved(final long currentNanos) {
+      if (acquired && referenceCount == 0 && currentNanos - releasedNanos > keepAliveNanos) {
```
stIncMale, line 168: While it's true that the `acquired` check would have eliminated the race condition that would have allowed the method `MongoClientCache.acquire` to return a closed ` MongoClient`, the `MongoClientCache` as it is currently written does not have this race condition anyway because executions of the methods `MongoClientCache.acquire` and `MongoClientCache.checkClientCache` are sequential (both methods are `synchronized`).

The `acquired` field serves no purposes even if we say that `MongoClientCacheItem` must not rely on the knowledge of how it is used (which is a valid design decision, but the alternative is also valid: it may rely on that knowledge as it's a `private` static class nested in `MongoClientCache`), because the `MongoClientCacheItem` is not thread-safe, and, thus, already relies on the correct synchronization provided by `MongoClientCache`.

stIncMale (changes requested):

src/main/java/com/mongodb/spark/sql/connector/connection/MongoClientCache.java
stIncMale: I understand this, and I am OK with the linear algorithmic complexity of the `MongoClientCache.release` method. It could have been made constant-time if it were to use the same releasing approach that is used in `DefaultConnectionPool`: decorate `MongoClient` such that its `close` method calls `MongoClientCacheItem.release()` and does not call `super.close()`. But this approach requires writing more code as it requires to implement a simple `MongoClient` decorator.

But I still do not understand why `ArrayList` is used instead of `HashMap` to hold cached objects. I tried to find a single benefit this provides, and I failed. This approach has, however, multiple drawbacks:

1. This approach requires writing more code because instead of simply doing `return map.computeIfAbsent(mongoClientFactory, () -> new MongoClientCacheItem(mongoClientFactory, keepAliveNanos)).acquire()`, the method `MongoClientCache.acquire` has to find an element in a list and modify the list if the element is not there, which results in 11 lines of code where 1 would be enough if `HashMap` were used.
2. This approach is surprising to a reader and makes it think hard about the reasons `ArrayList` was chosen over `HashMap`. For me this thinking was futile.
3. This approach results in higher algorithmic complexity. Since this is irrelevant for the expected use cases, having even a small benefit of preferring the current approach could potentially outweigh this specific downside, but I failed to find a single benefit of using `ArrayList` instead of `HashMap`.

Could you please help me to understand what are the benefits of using `ArrayList` here instead of `HashMap`?

src/main/java/com/mongodb/spark/sql/connector/config/Configurable.java
```diff
@@ -0,0 +1,29 @@
...
+
+  /**
+   * Configure this class with the {@link MongoConfig}.
+   *
+   * @param config the current configuration
+   */
+  void configure(MongoConfig config);
```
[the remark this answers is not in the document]
stIncMale: So `Configurable.configure` is similar to a method annotated with [`@PostConstruct`](https://docs.oracle.com/javase/8/docs/api/javax/annotation/PostConstruct.html) (while it is part of the Java SE 8 API, it was later moved to [Jakarta Annotations](https://jakarta.ee/specifications/annotations/2.0/apidocs/jakarta.annotation/jakarta/annotation/postconstruct)).

> Users can provide names of classes
> allows users to pass configurable instances

In the current PR I see only the former, where a class name may be specified and mongo-spark creates its instance and calls `Configurable.configure` on it. Was the latter a typo, or are we actually expecting to have an API that allows users to pass `Configurable` objects (not names of classes) to mongo-spark, where we will inject a configuration by calling `configure`?

The above is an important question, because if it is always the case that mongo-spark creates an instance of `Configurable` and calls `configure` before using the instance, then the `Configurable` interface only complicates matters without providing any benefit. In this case we can instead say that the classes users specify must have either a `public` constructor that takes no parameters, or a `public` constructor that takes a single parameter of type `MongoConfig`.

PR #59, opened 2021-11-23
Implement Spark read API

rozza:
SPARK-201


src/main/java/com/mongodb/spark/sql/connector/read/MongoPartitionReader.java
```diff
@@ -0,0 +1,119 @@
...
+  public boolean next() {
+    Assertions.ensureState(() -> !closed, () -> "Cannot call next() on a closed PartitionReader.");
+    return getCursor().hasNext();
+  }
+
+  /** Return the current record. This method should return same value until `next` is called. */
+  @Override
+  public InternalRow get() {
+    Assertions.ensureState(() -> !closed, () -> "Cannot call get() on a closed PartitionReader.");
+    return bsonDocumentToRowConverter.toInternalRow(getCursor().next());
```
stIncMale, line 81: I think the current implementation has two problems:

1. `PartitionReader.get` [_"should return same value until `next` is called"_](https://spark.apache.org/docs/3.1.2/api/java/org/apache/spark/sql/connector/read/PartitionReader.html#get--). This requirement is violated.
2. The fact that `PartitionReader.next` is declared as `throws IOException`, but `PartitionReader.get` is not, implies that `PartitionReader.next` must do the IO, while `PartitionReader.get` should only return the data read by `PartitionReader.next`.

src/main/java/com/mongodb/spark/sql/connector/read/MongoPartitionReader.java
```diff
@@ -0,0 +1,119 @@
...
+    return mongoCursor;
+  }
+
+  private void releaseCursor() {
+    if (mongoCursor != null) {
+      LOGGER.debug("Closing cursor for partitionId: {}", partition.getPartitionId());
+      mongoCursor.close();
+      mongoCursor = null;
+      mongoClient.close();
+      mongoClient = null;
```
stIncMale, line 116: Closing resources this way instead of using the `try`-`finally` or `try`-with-resources statements is prone to resource leaks: if `mongoCursor.close()` completes abruptly, `mongoClient.close()` is not executed.

stIncMale (changes requested):

src/main/java/com/mongodb/spark/sql/connector/read/MongoPartitionReader.java
rozza: Done

src/main/java/com/mongodb/spark/sql/connector/read/MongoPartitionReader.java
rozza: Good catch - done

src/main/java/com/mongodb/spark/sql/connector/read/package-info.java
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+/** This package contains the sql connector write support. */
```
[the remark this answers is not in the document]
rozza: Done

src/main/java/com/mongodb/spark/sql/connector/schema/JsonWriterSettingsSupplier.java
```diff
@@ -0,0 +1,30 @@
...
+import java.util.function.Supplier;
+
+import org.bson.json.JsonWriterSettings;
+
+/**
+ * A concrete supplier of JsonWriterSettings that is serializable.
+ *
+ * <p>Concrete implementations are used to convert {@code BsonDocument} into a {@code Row}.
+ */
+public interface JsonWriterSettingsSupplier extends Supplier<JsonWriterSettings>, Serializable {}
```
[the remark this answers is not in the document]
rozza: I'm going to keep this around - users will want to control how bson types are represented as strings.

src/main/java/com/mongodb/spark/sql/connector/schema/ConverterHelper.java
```diff
@@ -47,5 +49,13 @@
           .symbolConverter((value, writer) -> writer.writeString(value))
           .build();
 
+  static class DefaultJsonWriterSettingsSupplier
+      implements Supplier<JsonWriterSettings>, Serializable {
```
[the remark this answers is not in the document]
rozza: Done

src/main/java/com/mongodb/spark/sql/connector/schema/BsonDocumentToRowConverter.java
```diff
@@ -85,46 +89,79 @@
...
+   * Converts a {@code BsonDocument} into a {@link GenericRowWithSchema} using the supplied {@code
    * StructType}
    *
    * @param bsonDocument the bson document to shape into a GenericRowWithSchema.
    * @return the converted data as a Row
    */
-  public Row fromBsonDocument(final BsonDocument bsonDocument) {
-    return fromBsonDocument((StructType) getDataType(bsonDocument), bsonDocument);
+  public Row toRow(final BsonDocument bsonDocument) {
+    return convertToRow("", schema, bsonDocument, jsonWriterSettingsSupplier.get());
```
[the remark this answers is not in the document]
rozza: While extended json has the value of being able to round trip data, users don't necessarily want to use it for string representations of data. Eg they often prefer ObjectId hex strings or human readable dates.

As Users typically don't agree with our json representations of values, we will likely add a configuration to allow users to provide their own or provide a choice of defaults.

src/main/java/com/mongodb/spark/sql/connector/schema/JsonWriterSettingsSupplier.java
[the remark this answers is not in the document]
rozza: Scratch that - we can resurrect the interface if and when needed.

src/main/java/com/mongodb/spark/sql/connector/schema/BsonDocumentToRowConverter.java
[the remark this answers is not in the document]
rozza: Done - removed the supplier for now and we can look to add in the future when the configuration is added.

src/main/java/com/mongodb/spark/sql/connector/read/MongoPartitionReader.java
rozza: Is there a nicer way than the nested try's?

stIncMale: Yes, the `try`-with-resources statement can execute a sequence of actions expressed as `AutoCloseable` objects in a way similar to nested `try`-`finally` statements without nesting.

P.S. Irrelevant for our case, and relevant only in situations when the number of actions is not known at compile time: Guava has [`Closer](https://guava.dev/releases/19.0/api/docs/com/google/common/io/Closer.html) that may execute a dynamic number of actions with behavior similar to that of the `try`-with-resources statement.

src/main/java/com/mongodb/spark/sql/connector/read/MongoPartitionReader.java
stIncMale: `get` now returns the same value until `next` is called, however, the `next` method does not necessary ["_Proceed to next record_"](https://spark.apache.org/docs/3.1.2/api/java/org/apache/spark/sql/connector/read/PartitionReader.html#next--). If we are going to base the implementation of `MongoPartitionReader.get` and `MongoPartitionReader.next` on the documentation of the methods `MongoCursor.hasNext`/`MongoCursor.next`<sup>1</sup>, then in order to satisfy the specification of the method `PartitionReader.next` we have to

- call `MongoCursor.next` in `MongoPartitionReader.next` if `MongoCursor.hasNext` produces `true`;
- store the result of `MongoCursor.next` for later use by `MongoPartitionReader.get`.

---
<sup>1</sup> As we should, unless there is a good reasons to break encapsulation and start depending on the undocumented behavior.

rozza: Updated.

PR #60, opened 2021-12-13
Infer the schema

rozza:
Determine the schema from a BsonDocument
Determine the common schema for a collection of BsonDocuments

SPARK-298


src/main/java/com/mongodb/spark/sql/connector/schema/InferSchema.java
```diff
@@ -0,0 +1,348 @@
...
+    switch (bsonValue.getBsonType()) {
+      case DOCUMENT:
+        List<StructField> fields = new ArrayList<>();
+        bsonValue
+            .asDocument()
+            .forEach(
+                (k, v) ->
+                    fields.add(
+                        new StructField(k, getDataType(v, readConfig), true, Metadata.empty())));
+        return dataTypeCheckStructTypeToMapType(DataTypes.createStructType(fields), readConfig);
```
stIncMale, line 110: Just a question: why do we represent some documents as `StructType` and some as `MapType` depending on how many fields (`INFER_SCHEMA_MAP_TYPE_MINIMUM_KEY_SIZE_CONFIG`) the document has?

stIncMale (approved):
There is a typo in a Javadoc, an a question, but othewise feel free to merge.

P.S. I am not pretending I fully understand all the logic in `InferSchema`, but it looks nice)

src/main/java/com/mongodb/spark/sql/connector/config/AbstractMongoConfig.java
```diff
@@ -148,7 +148,7 @@ public MongoClient getMongoClient() {
   }
 
   /**
-   * Loans a {@link MongoClient} to the user.
+   * Runs a function against a {@code MongoClient}
```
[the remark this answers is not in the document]
rozza: I thought that when looking at the the diff - it makes it look like its for `doWithCollection` but its actually the javadoc for `withClient`

src/main/java/com/mongodb/spark/sql/connector/schema/InferSchema.java
rozza: Good question - this feature was added as a user request in the original connector.  

Their data was very dynamic and had many different key names. So trying to merge them into a single StructType was expensive in the size of the Structs and error prone as the sampling wouldn't pick up all the variations in their map like data.

src/main/java/com/mongodb/spark/sql/connector/config/AbstractMongoConfig.java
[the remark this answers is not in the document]
stIncMale: Ah, right!

src/main/java/com/mongodb/spark/sql/connector/schema/InferSchema.java
stIncMale: So `MapType` is more versatile than `StructType`, got it.

PR #61, opened 2022-01-06
Spark streams can call commit with no data

rozza:
Prevents the mongoCollection.bulkWrite throwing an error
on an empty batch.

SPARK-202


src/integrationTest/java/com/mongodb/spark/sql/connector/write/MongoSparkConnectorWriteTest.java
```diff
@@ -89,6 +93,44 @@ void testSupportedStreamingWriteModes() throws TimeoutException {
     assertCollection();
   }
 
+  @Test
+  void testSupportedStreamingWriteComplete() throws TimeoutException {
```
stIncMale, line 97: My guess is that this test causes `MongoDataWriter.commit` to be called when `MongoDataWriter.writeModelList` is empty, but I do not understand what in the test causes that to happen. If my guess is correct, could you please explain this via a comment in the test? If the guess is incorrect, then what is the relation between the fix in this PR and the test?

rozza: Done.

stIncMale: I'd appreciate if we could have a call before/after the standup meeting and you just walk me through the test and how it causes `MongoDataWriter.commit` to be called when `MongoDataWriter.writeModelList` is empty. The comment does not shed any light on this.

stIncMale (approved):

PR #65, opened 2022-02-11
Support Continuous streams using change streams

rozza:
SPARK-205

### DIFF
--- a/src/main/java/com/mongodb/spark/sql/connector/read/MongoContinuousStream.java
+++ b/src/main/java/com/mongodb/spark/sql/connector/read/MongoContinuousStream.java
@@ -37,6 +37,9 @@ import com.mongodb.spark.sql.connector.schema.BsonDocumentToRowConverter;
 /**
  * MongoContinuousStream defines how to read a stream of data from MongoDB.
  *
+ * <p>Utilizes MongoDBs change stream functionality,  the continuous streams will consist of
+ * <a href="https://docs.mongodb.com/manual/reference/change-events/">change events</a>.
+ *
  * <p>Note: Requires MongoDB 4.2+ To support continuing a change stream after a collection has been dropped.
  **/
 public class MongoContinuousStream implements ContinuousStream {
--- a/src/main/java/com/mongodb/spark/sql/connector/read/MongoContinuousStream.java
+++ b/src/main/java/com/mongodb/spark/sql/connector/read/MongoContinuousStream.java
@@ -1,0 +1,105 @@
+/*
+ * Copyright 2008-present MongoDB, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package com.mongodb.spark.sql.connector.read;
+
+import static com.mongodb.spark.sql.connector.read.ResumeTokenOffset.INITIAL_RESUME_TOKEN_OFFSET;
+import static java.lang.String.format;
+import static java.util.Collections.emptyList;
+
+import org.apache.spark.sql.connector.read.InputPartition;
+import org.apache.spark.sql.connector.read.streaming.ContinuousPartitionReaderFactory;
+import org.apache.spark.sql.connector.read.streaming.ContinuousStream;
+import org.apache.spark.sql.connector.read.streaming.Offset;
+import org.apache.spark.sql.connector.read.streaming.PartitionOffset;
+import org.apache.spark.sql.types.StructType;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.mongodb.spark.sql.connector.assertions.Assertions;
+import com.mongodb.spark.sql.connector.config.ReadConfig;
+import com.mongodb.spark.sql.connector.schema.BsonDocumentToRowConverter;
+
+/** MongoContinuousStream defines how to read a stream of data from MongoDB. */
+public class MongoContinuousStream implements ContinuousStream {
+  private static final Logger LOGGER = LoggerFactory.getLogger(MongoContinuousStream.class);
+  private final BsonDocumentToRowConverter bsonDocumentToRowConverter;
+  private final ReadConfig readConfig;
+  private int partitionId = -1;
+
+  /**
+   * Construct a new instance
+   *
+   * @param schema the schema for the data
+   * @param readConfig the read configuration
+   */
+  public MongoContinuousStream(final StructType schema, final ReadConfig readConfig) {
+    Assertions.validateConfig(
+        schema,
+        (s) -> !s.isEmpty(),
+        () -> "Mongo Continuous streams require a schema to be defined");
+    this.bsonDocumentToRowConverter = new BsonDocumentToRowConverter(schema);
+    this.readConfig = readConfig;
+  }
+
+  @Override
+  public InputPartition[] planInputPartitions(final Offset start) {
+    this.partitionId += 1;
+    return new InputPartition[] {
+      new MongoInputPartition(this.partitionId, emptyList(), (ResumeTokenOffset) start)
+    };
+  }
+
+  @Override
+  public ContinuousPartitionReaderFactory createContinuousReaderFactory() {
+    return new MongoStreamPartitionReaderFactory(bsonDocumentToRowConverter, readConfig);
+  }
+
+  @Override
+  public Offset mergeOffsets(final PartitionOffset[] offsets) {
+    Assertions.ensureState(
+        () -> offsets.length == 1, () -> "Multiple offsets found when there should only be one.");
+    Assertions.ensureState(
+        () -> offsets[0] instanceof ResumeTokenPartitionOffset,
+        () ->
+            format(
+                "Unexpected partition offset type. "
+                    + "Expected ResumeTokenPartitionOffset` found `%s`",
+                offsets[0].getClass()));
+    return new ResumeTokenOffset(((ResumeTokenPartitionOffset) offsets[0]).getResumeToken());
+  }
+
+  @Override
+  public Offset initialOffset() {
+    return INITIAL_RESUME_TOKEN_OFFSET;
+  }
+
+  @Override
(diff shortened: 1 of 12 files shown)


src/main/java/com/mongodb/spark/sql/connector/read/MongoStreamPartitionReader.java
```diff
@@ -0,0 +1,204 @@
...
+      LOGGER.info("Opened change stream cursor for partitionId: {}", partition.getPartitionId());
+      ChangeStreamIterable<Document> changeStreamIterable =
+          mongoClient
+              .getDatabase(readConfig.getDatabaseName())
+              .getCollection(readConfig.getCollectionName())
+              .watch(partition.getPipeline());
+
+      if (!lastOffset.getResumeToken().isEmpty()) {
+        changeStreamIterable = changeStreamIterable.resumeAfter(lastOffset.getResumeToken());
+      }
```
stIncMale, line 160: Is my understanding correct, that when the offset is empty and no new writes are happening, `changeStreamIterable` will not return any documents, and, therefore, `MongoStreamPartitionReader` won't read any data?

- If yes, then how can this reader read any data, given that the offset comes from the Spark framework via the `MongoContinuousStream.planInputPartitions` method, and Spark framework cannot pass any offset other than `MongoContinuousStream.initialOffset` unless it got the offset previously from `MongoContinuousStream.mergeOffsets` (merged offsets come from `MongoStreamPartitionReader.getOffset`)?
- If no, then it means that change streams return all changes from the "beginning of time" unless `resumeAfter`/`startAfter`/`startAtOperationTime` is specified, and I failed to find any mention of this behavior.

src/main/java/com/mongodb/spark/sql/connector/read/MongoStreamPartitionReader.java
```diff
@@ -0,0 +1,204 @@
...
+                .getValue()
+                .toLowerCase(Locale.ROOT)
+                .equals(INVALIDATE)) {
+              LOGGER.info(
+                  "Change stream cursor has been invalidated. This happens when a collection is dropped. Closing cursor.");
+              lastOffset = EMPTY_RESUME_TOKEN_PARTITION_OFFSET;
+              releaseCursor();
+              releaseClient();
+            }
+            currentRow = bsonDocumentToRowConverter.toInternalRow(next);
```
stIncMale, line 118: Instead of reading the actual documents here, we are reading [change events](https://docs.mongodb.com/manual/reference/change-events/#std-label-change-stream-output), which is different from what `MongoPartitionReader` reads. This, in turn, means that `MongoScan.toContinuousStream` and `MongoScan.toBatch` read change events vs actual documents. Is it not the case that they are supposed to read the same data but in different ways?

src/main/java/com/mongodb/spark/sql/connector/read/MongoContinuousStream.java
```diff
@@ -0,0 +1,105 @@
...
+        schema,
+        (s) -> !s.isEmpty(),
+        () -> "Mongo Continuous streams require a schema to be defined");
+    this.bsonDocumentToRowConverter = new BsonDocumentToRowConverter(schema);
+    this.readConfig = readConfig;
+  }
+
+  @Override
+  public InputPartition[] planInputPartitions(final Offset start) {
+    this.partitionId += 1;
```
stIncMale, line 61: Given that `MongoContinuousStream.planInputPartitions` always produces exactly one partition and is called not more than once per instance of `MongoStreamPartitionReader`, because `MongoContinuousStream.needsReconfiguration` always returns `false`, why do we have the `partitionId` field? It appears that `planInputPartitions` should just pass 0 as the partition ID when it creates an instance of `MongoInputPartition`.

stIncMale (changes requested):
My comments are a mix of questions that I was unable to find answers to on my own, and suggestions.

src/main/java/com/mongodb/spark/sql/connector/read/MongoContinuousStream.java
```diff
@@ -0,0 +1,105 @@
...
+    return new InputPartition[] {
+      new MongoInputPartition(this.partitionId, emptyList(), (ResumeTokenOffset) start)
```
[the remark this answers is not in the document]
rozza: Ack

src/main/java/com/mongodb/spark/sql/connector/read/MongoStreamPartitionReader.java
```diff
@@ -0,0 +1,204 @@
...
+
+    this.readConfig = readConfig;
+    this.currentRow = null;
+    this.lastOffset =
+        new ResumeTokenPartitionOffset(partition.getResumeTokenOffset().getResumeToken());
+
+    LOGGER.debug(
+        "Creating partition reader for: PartitionId: {} with Schema: {}",
+        partition.getPartitionId(),
+        bsonDocumentToRowConverter.getSchema());
```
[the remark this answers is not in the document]
rozza: SGTM

src/main/java/com/mongodb/spark/sql/connector/read/MongoStreamPartitionReader.java
```diff
@@ -0,0 +1,204 @@
...
+          return true;
+        });
+  }
+
+  @Override
+  public InternalRow get() {
+    Assertions.ensureState(() -> !closed, () -> "Cannot call get() on a closed PartitionReader.");
+    Assertions.ensureState(
+        () -> currentRow != null,
+        () -> "Current row is null, this should not happen if `next()` returns false.");
```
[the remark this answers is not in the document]
rozza: Thanks

src/main/java/com/mongodb/spark/sql/connector/read/MongoStreamPartitionReader.java
```diff
@@ -0,0 +1,204 @@
...
+      releaseClient();
+    }
+  }
+
+  private MongoChangeStreamCursor<BsonDocument> getCursor() {
+    if (mongoClient == null) {
+      mongoClient = readConfig.getMongoClient();
+    }
+    if (changeStreamCursor == null) {
+      LOGGER.info("Opened change stream cursor for partitionId: {}", partition.getPartitionId());
```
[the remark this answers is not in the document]
rozza: :+1:  Good catch

src/main/java/com/mongodb/spark/sql/connector/read/MongoStreamPartitionReader.java
```diff
@@ -0,0 +1,204 @@
...
+    LOGGER.trace("Get called: {}", currentRow);
+    return currentRow;
+  }
+
+  @Override
+  public void close() {
+    if (!closed) {
+      closed = true;
+      releaseCursor();
+      releaseClient();
```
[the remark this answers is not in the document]
rozza: Ack

src/main/java/com/mongodb/spark/sql/connector/read/MongoStreamPartitionReader.java
```diff
@@ -0,0 +1,204 @@
...
+          try {
+            BsonDocument next = c.next();
+            lastOffset = new ResumeTokenPartitionOffset(c.getResumeToken());
+            if (next.getString("operationType", new BsonString(""))
+                .getValue()
+                .toLowerCase(Locale.ROOT)
+                .equals(INVALIDATE)) {
+              LOGGER.info(
+                  "Change stream cursor has been invalidated. This happens when a collection is dropped. Closing cursor.");
+              lastOffset = EMPTY_RESUME_TOKEN_PARTITION_OFFSET;
```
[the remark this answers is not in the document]
rozza: So this is tricky. An invalidated change stream event doesn't mean there wont be more changes on the collection in the future. Its just that invalidate events happen to kill the current cursor and requires a new cursor.

Why use an empty resume token? this is a hang over from pre MongoDB 4.2 where users would have to use resumeAfter - however, if we limit users to MongoDB 4.2+ then startAfter is the best api. So I have updated to use that. It will stop a data race and potential loss of data.

src/main/java/com/mongodb/spark/sql/connector/read/MongoStreamPartitionReader.java
rozza: Both `MongoPartitionReader` and `MongoStreamPartitionReader` read bson documents as change events are just bson documents. So both partitioners read bson documents and convert them to output InternalRow data.

rozza:
Thanks @stIncMale all comments should have been actioned or replied to :)

src/main/java/com/mongodb/spark/sql/connector/read/MongoStreamPartitionReader.java
stIncMale: Thanks. So at the end of the day, the data read by `MongoStreamPartitionReader` is determined by the `Offset` passed to `ContinuousStream.planInputPartitions`. There is no concept of reading everything in `ContinuousStream`, and therefore it's fine that we cannot read "everything" via the change stream either.

src/main/java/com/mongodb/spark/sql/connector/read/MongoStreamPartitionReader.java
stIncMale: Sure, they both read `BsonDocument`. The difference I am talking about is in what this document is: in case of `MongoPartitionReader` the document is what an aggregation pipeline outputs, while in case of `MongoStreamPartitionReader` the document is a change event. My expectation would be that `Scan.toBatch` and `Scan.toContinuousStream` allow me to read the same documents, and that's not the case in Mongo Spark. Is my expectation groundless and users don't expect this?

Either way, it seems worth at the very least documenting (including javadocs) that `MongoScan.toBatch` and `MongoScan.toContinuousStream` read different documents: the former reads documents outputted by an aggregation pipeline (most likely simply documents stored in a collection), while the latter reads change events.

stIncMale (commented):
.

src/main/java/com/mongodb/spark/sql/connector/read/MongoStreamPartitionReader.java
rozza: Well to be devils advocate change events are just aggregation results also ;) but I get your point - will clarify.

Will clarify the java doc - but the real place this needs documenting (as users wont see the java doc) is in the usage documentation.  Also there will be an option to just output full documents only in another ticket - but change events are more than just insert events.

src/main/java/com/mongodb/spark/sql/connector/read/MongoStreamPartitionReader.java
rozza: Correct.

rozza:
Updated the java doc @stIncMale

stIncMale (approved):

PR #66, opened 2022-03-01
SPARK-324 - MongoSpark SAVE API which accepts MongoConnector instance as argument

praneetsharma:
PR for SPARK-324: MongoSpark SAVE API which accepts MongoConnector instance as argument

- Added a new save method in MongoSpark.scala which also accepts MongoConnector instance.
- Increased maxMethods count to 35 since MongoSpark now has over 30 methods.


rozza:
Thanks @praneetsharma

PR #69, opened 2022-04-25
SPARK-343: More save methods in MongoSpark accepting MongoConnector as argument

praneetsharma:
SPARK-343: More save methods in MongoSpark accepting MongoConnector as argument


rozza:
Hi @praneetsharma,

There are no plans to further extend the Scala API as development has moved towards the 10.0.x version of the connector.

Ross.

PR #70, opened 2022-04-28
Ensure converting bson types produces valid types

rozza:
Use Timestamp instead of Date
Use Arrays instead of ArrayLists
Handle BigDecimal as well as Decimal

SPARK-340 SPARK-342


jyemin (commented):
Changes look good.  I'm just not sure if either of the type changes will cause compatibility issues when applications upgrade. Would any existing applications be expecting Date or List instead of Timestamp or Object array?

rozza:
The changes were needed due to how Spark returns / provides types. So essentially we were providing the wrong types for the scala type.

jyemin (approved):
LGTM

PR #71, opened 2022-05-04
Pass the users pipeline to inferSchema

rozza:
SPARK-344


jyemin (approved):
Looks good to me.  Is there anything else we might have missed that is similar?

PR #72, opened 2022-05-13
[SPARK-346] Fix convert date type to bson

tsekityam:
When I try to write a databricks data frame with `DateType` to mongodb, I got `Error: Cannot cast 2022-05-13 into a BsonValue. DateType has no matching BsonValue. Error: java.sql.Date cannot be cast to java.sql.Timestamp`

I found that the connector treats `DateType` as `jave.sql.Timestamp` during conversion, however, `DateType` should be `jave.sql.Date` and `jave.sql.Date` can't be casted as `jave.sql.Timestamp`

This PR fixes the `DateType` to bson conversion, and also adds a test case to ensure the conversion works as expected.


rozza:
Hi @tsekityam just to let you know this was fixed and is available in the latest version of the connector 10.0.2

PR #75, opened 2022-07-04
ShardedPartitioner improvements

rozza:
Throw an error if the user has a hashed or compound shard key

SPARK-345

### DIFF
--- a/src/main/java/com/mongodb/spark/sql/connector/read/partitioner/ShardedPartitioner.java
+++ b/src/main/java/com/mongodb/spark/sql/connector/read/partitioner/ShardedPartitioner.java
@@ -109,15 +109,9 @@ public final class ShardedPartitioner implements Partitioner {
 
     BsonDocument keyDocument = configCollectionMetadata.getDocument("key", new BsonDocument());
     if (keyDocument.keySet().size() > 1) {
-      LOGGER.warn(
-          "Compound shard key: '{}'. Not supported, please use an alternative partitioner.",
-          keyDocument.toJson());
       throw new MongoSparkException(
           "Invalid partitioner strategy. The Sharded partitioner does not support compound shard keys.");
     } else if (keyDocument.containsValue(new BsonString("hashed"))) {
-      LOGGER.warn(
-          "Hashed shard key: '{}'. Not supported, please use an alternative partitioner.",
-          keyDocument.toJson());
       throw new MongoSparkException(
           "Invalid partitioner strategy. The Sharded partitioner does not support hashed shard keys.");
     }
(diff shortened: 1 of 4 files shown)


src/main/java/com/mongodb/spark/sql/connector/read/partitioner/ShardedPartitioner.java
```diff
@@ -105,6 +107,21 @@ public List<MongoInputPartition> generatePartitions(final ReadConfig readConfig)
...
+          "Compound shard key: '{}'. Not supported, please use an alternative partitioner.",
+          keyDocument.toJson());
+      throw new MongoSparkException(
+          "Invalid partitioner strategy. The Sharded partitioner does not support compound shard keys.");
+    } else if (keyDocument.containsValue(new BsonString("hashed"))) {
+      LOGGER.warn(
+          "Hashed shard key: '{}'. Not supported, please use an alternative partitioner.",
+          keyDocument.toJson());
+      throw new MongoSparkException(
+          "Invalid partitioner strategy. The Sharded partitioner does not support hashed shard keys.");
```
stIncMale, line 122: Why do we need to both log and throw here? `FieldPartitioner.createMongoInputPartitions` throws and does not duplicate the error message in logs.

If we indeed need to duplicate messages, should these messages be `ERROR`, given that the connector actually fails?

src/main/java/com/mongodb/spark/sql/connector/read/partitioner/ShardedPartitioner.java
```diff
@@ -54,7 +55,8 @@
  * <p>Uses the chunks collection and partitions the collection based on the sharded collections
  * chunk ranges.
  *
- * <p><strong>Note:</strong> Does not support collections sharded using hashed shard keys.
+ * <p><strong>Note:</strong> Does not support collections sharded using hashed shard keys or
+ * compound shard keys.
```
stIncMale, line 59: [just a comment]
Indeed, this is something we missed in the original [discussion](https://github.com/mongodb/mongo-spark/pull/63#discussion_r813419181). If [`$lt`](https://www.mongodb.com/docs/manual/reference/operator/query/lt/)&Co were to support comparing a compound "field" against a compound value (which is what the server does when computes chunks and decides which chunk a document goes into, and what `$sort` does, so technically this appear to be possible), then `ShardedPartitioner` and all other partitioners could support compound keys. It is worth noting that other partitioners would, unlike `ShardedPartitioner`, still not be able to support non-unique partition keys.

src/main/java/com/mongodb/spark/sql/connector/read/partitioner/ShardedPartitioner.java
rozza: Unlike the other loggings where a single partitioner is used instead, here we error. So we don't need the logs at all.

Removed.

stIncMale (approved):

PR #77, opened 2022-08-17
Updated MongoScanBuilder

rozza:
It now handles error cases better by not throwing an exception. It now
will leave the Filter for Spark to handle.

Now handles nested lookups and pushes them down to the aggregation
framework.

SPARK-359

### DIFF
--- a/src/main/java/com/mongodb/spark/sql/connector/read/MongoScanBuilder.java
+++ b/src/main/java/com/mongodb/spark/sql/connector/read/MongoScanBuilder.java
@@ -25,6 +25,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Locale;
+import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
 
@@ -48,11 +49,13 @@ import org.apache.spark.sql.sources.Or;
 import org.apache.spark.sql.sources.StringContains;
 import org.apache.spark.sql.sources.StringEndsWith;
 import org.apache.spark.sql.sources.StringStartsWith;
+import org.apache.spark.sql.types.DataType;
 import org.apache.spark.sql.types.StructField;
 import org.apache.spark.sql.types.StructType;
 import org.jetbrains.annotations.Nullable;
 
 import org.bson.BsonDocument;
+import org.bson.BsonValue;
 import org.bson.conversions.Bson;
 
 import com.mongodb.client.model.Aggregates;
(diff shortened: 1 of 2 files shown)


src/main/java/com/mongodb/spark/sql/connector/read/MongoScanBuilder.java
```diff
@@ -194,35 +197,44 @@ private FilterAndPipelineStage processFilter(final Filter filter) {
...
+              .orElse(null));
     } else if (filter instanceof In) {
       In inFilter = (In) filter;
-      Bson pipelineStage =
-          Filters.in(
-              inFilter.attribute(),
-              Arrays.stream(inFilter.values())
-                  .map(v -> processValue(inFilter.attribute(), v))
-                  .collect(Collectors.toList()));
+      List<BsonValue> values =
```
jyemin, line 226: From what I can tell by running the modified test with a coverage tool, this branch is not tested (nor are most of the modified branches above).  Can the test coverage be improved?

jyemin (changes requested):

src/main/java/com/mongodb/spark/sql/connector/read/MongoScanBuilder.java
rozza: There is a test for all the filters: MongoSparkConnectorReadTest#testReadsAreSupportedWithFilters that remains unchanged.

Although I can't get the AND filter to be picked explicitly Spark, it coming as multiple Filters in DataSourceV2. But the logic for it seems sound so I don't want to remove it incase Spark gets updated in the future.

jyemin: My concern is that the modified test doesn't cover the modified code paths, which implies that there are no tests for the new behavior.   Sorry if I'm misunderstanding.

rozza: The pipeline stage resulting is the same for the IN test we currently have. The extra logic is belt and braces as I don't know how to write an erroneous IN column that would force us to not be able to parse the BsonValue. 

That said sometimes users have used Scala types that have cause issues before. In the event of that happening I only want to push down the IN statement if all values can be converted to BsonValues - otherwise it would potentially miss data.

jyemin: OK, resolving.

jyemin (approved):

PR #79, opened 2022-08-23
Ensure MongoClient#getOptions returns case insensitive options

rozza:
SPARK-363

### DIFF
--- a/src/integrationTest/java/com/mongodb/spark/sql/connector/read/MongoSparkConnectorReadTest.java
+++ b/src/integrationTest/java/com/mongodb/spark/sql/connector/read/MongoSparkConnectorReadTest.java
@@ -27,6 +27,7 @@ import static org.junit.jupiter.api.Assertions.assertInstanceOf;
 import static org.junit.jupiter.api.Assertions.assertIterableEquals;
 import static org.junit.jupiter.api.Assertions.assertNull;
 import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.junit.jupiter.api.Assertions.fail;
 import static org.junit.jupiter.api.Assumptions.assumeTrue;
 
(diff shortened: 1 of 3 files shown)


src/integrationTest/java/com/mongodb/spark/sql/connector/read/MongoSparkConnectorReadTest.java
```diff
@@ -442,6 +443,26 @@ void testReadsAreSupportedWithFilters() {
...
+
+    List<BsonDocument> collectionData =
+        toBsonDocuments(spark.read().textFile(READ_RESOURCES_HOBBITS_JSON_PATH));
+    getCollection().insertMany(collectionData);
+
+    spark
+        .read()
+        .format("mongodb")
+        .option(
+            ReadConfig.READ_PREFIX + ReadConfig.CLIENT_FACTORY_CONFIG,
```
rozza, line 458: Here we test setting an option that previously would not have been picked up after the ReadConfig options had been specialized due to it being converted to lowercase.

src/main/java/com/mongodb/spark/sql/connector/config/AbstractMongoConfig.java
```diff
@@ -70,6 +69,7 @@ enum UsageMode {
   private final Map<String, String> options;
   private final UsageMode usageMode;
   private transient MongoClientFactory mongoClientFactory;
+  private transient CaseInsensitiveStringMap caseInsensitiveOptions;
```
rozza, line 72: Unfortunately, CaseInsensitiveStringMap is not serializable - so we make it transient

src/integrationTest/java/com/mongodb/spark/sql/connector/read/MongoSparkConnectorReadTest.java
jyemin: Ack

jyemin (approved):

PR #80, opened 2022-09-23
New flag 'upsertDocument' to control the insert behavior of writes

santhoshsuresh95:
'upsertDocument' flag passed as spark option control the insert behavior (provided if 'forceInsert' is set to false)

If true - MongoSpark behavior the same as earlier i.e. inserts / updates when saving a Dataset.
If false - MongoSpark does not insert new documents and updates / sets fields for matching documents.


rozza:
There are no plans for further 3.0.x releases.

This fix is being ported to the 10.1.0 release.

PR #83, opened 2022-11-03
Added MicroBatch Stream support

rozza:
SPARK-368

### DIFF
--- a/src/integrationTest/java/com/mongodb/spark/sql/connector/read/AbstractMongoStreamTest.java
+++ b/src/integrationTest/java/com/mongodb/spark/sql/connector/read/AbstractMongoStreamTest.java
@@ -1,0 +1,386 @@
+/*
+ * Copyright 2008-present MongoDB, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+package com.mongodb.spark.sql.connector.read;
+
+import static java.lang.String.format;
+import static java.util.Arrays.asList;
+import static org.apache.spark.sql.types.DataTypes.createStructField;
+import static org.apache.spark.sql.types.DataTypes.createStructType;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.junit.jupiter.api.Assertions.fail;
+import static org.junit.jupiter.api.Assumptions.assumeTrue;
+
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.TimeoutException;
+import java.util.function.Consumer;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+
+import org.apache.spark.sql.Column;
+import org.apache.spark.sql.Dataset;
+import org.apache.spark.sql.Row;
+import org.apache.spark.sql.SparkSession;
+import org.apache.spark.sql.streaming.StreamingQuery;
+import org.apache.spark.sql.streaming.StreamingQueryException;
+import org.apache.spark.sql.streaming.Trigger;
+import org.apache.spark.sql.types.DataTypes;
+import org.apache.spark.sql.types.StructType;
+import org.jetbrains.annotations.Nullable;
+import org.junit.jupiter.api.Test;
+
+import org.bson.BsonDocument;
+import org.bson.BsonInt32;
+import org.bson.BsonString;
+
+import com.mongodb.client.MongoCollection;
+import com.mongodb.client.model.Filters;
+import com.mongodb.client.model.Updates;
+
+import com.mongodb.spark.sql.connector.config.MongoConfig;
+import com.mongodb.spark.sql.connector.config.ReadConfig;
+import com.mongodb.spark.sql.connector.config.WriteConfig;
+import com.mongodb.spark.sql.connector.exceptions.ConfigException;
+import com.mongodb.spark.sql.connector.mongodb.MongoSparkConnectorTestCase;
+
+abstract class AbstractMongoStreamTest extends MongoSparkConnectorTestCase {
+
+  abstract String collectionPrefix();
+
+  abstract Trigger getTrigger();
+
+  @Test
+  void testStream() {
+    assumeTrue(supportsChangeStreams());
+    testStreamingQuery(
+        createMongoConfig("Simple"),
+        withSource(
+            coll ->
+                coll.insertMany(
+                    IntStream.range(0, 25)
+                        .mapToObj(i -> new BsonDocument("_id", new BsonInt32(i)))
+                        .collect(Collectors.toList()))),
+        withSink(coll -> assertEquals(25, coll.countDocuments(), "Expected to see 25 documents")),
+        withSource(
+            coll ->
+                coll.insertMany(
+                    IntStream.range(100, 125)
+                        .mapToObj(i -> new BsonDocument("_id", new BsonInt32(i)))
+                        .collect(Collectors.toList()))),
+        withSink(coll -> assertEquals(50, coll.countDocuments(), "Expected to see 50 documents")));
+  }
+
+  @Test
+  void testStreamHandlesCollectionDrop() {
+    assumeTrue(supportsChangeStreams());
+    assumeTrue(isAtLeastFourDotFour());
+
+    testStreamingQuery(
+        createMongoConfig("WithDrop"),
+        withSource(
+            coll ->
+                coll.insertMany(
+                    IntStream.range(0, 25)
+                        .mapToObj(i -> new BsonDocument("_id", new BsonInt32(i)))
+                        .collect(Collectors.toList()))),
+        withSink(coll -> assertEquals(25, coll.countDocuments(), "Expected to see 25 documents")),
+        withSource(MongoCollection::drop),
+        withSink(
+            coll ->
+                assertEquals(
+                    1,
+                    coll.countDocuments(Filters.eq("operationType", "drop")),
+                    "Expected to see 1 drop document")),
+        withSink(
+            coll ->
+                assertEquals(
+                    1,
+                    coll.countDocuments(Filters.eq("operationType", "invalidate")),
+                    "Expected to see 1 invalidate document")));
+  }
+
+  @Test
+  void testStreamWithFilter() {
+    assumeTrue(supportsChangeStreams());
+    assumeTrue(isAtLeastFourDotFour());
+
+    Column filterColumn = new Column("operationType").equalTo("insert");
+    testStreamingQuery(
+        createMongoConfig("WithFilter")
+            .withOption(
+                ReadConfig.READ_PREFIX + ReadConfig.STREAM_LOOKUP_FULL_DOCUMENT_CONFIG,
+                "updateLookup"),
+        DEFAULT_SCHEMA,
+        filterColumn,
+        withSource(
+            coll -> {
+              coll.insertMany(
+                  IntStream.range(0, 50)
+                      .mapToObj(i -> new BsonDocument("_id", new BsonInt32(i)))
+                      .collect(Collectors.toList()));
+              coll.deleteMany(
+                  Filters.in(
+                      "_id",
+                      IntStream.range(0, 50)
+                          .filter(i -> i % 2 == 0)
+                          .boxed()
+                          .collect(Collectors.toList())));
+            }),
+        withSink(coll -> assertEquals(50, coll.countDocuments(), "Seen 50 documents")));
+  }
+
+  @Test
+  void testStreamWithPublishFullDocumentOnly() {
+    assumeTrue(supportsChangeStreams());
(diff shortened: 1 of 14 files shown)


src/integrationTest/java/com/mongodb/spark/sql/connector/read/AbstractMongoStreamTest.java
```diff
@@ -0,0 +1,386 @@
...
+import com.mongodb.spark.sql.connector.config.ReadConfig;
+import com.mongodb.spark.sql.connector.config.WriteConfig;
+import com.mongodb.spark.sql.connector.exceptions.ConfigException;
+import com.mongodb.spark.sql.connector.mongodb.MongoSparkConnectorTestCase;
+
+abstract class AbstractMongoStreamTest extends MongoSparkConnectorTestCase {
+
+  abstract String collectionPrefix();
+
+  abstract Trigger getTrigger();
```
rozza, line 68: The trigger determines if its a Continuous stream or a Micro Batch stream

src/integrationTest/java/com/mongodb/spark/sql/connector/read/MongoBatchTest.java
```diff
@@ -48,29 +36,17 @@
...
-
-import com.mongodb.spark.sql.connector.config.MongoConfig;
 import com.mongodb.spark.sql.connector.config.ReadConfig;
-import com.mongodb.spark.sql.connector.config.WriteConfig;
-import com.mongodb.spark.sql.connector.exceptions.ConfigException;
 import com.mongodb.spark.sql.connector.mongodb.MongoSparkConnectorTestCase;
 import com.mongodb.spark.sql.connector.schema.RowToBsonDocumentConverter;
 
-class MongoSparkConnectorReadTest extends MongoSparkConnectorTestCase {
+class MongoBatchTest extends MongoSparkConnectorTestCase {
```
rozza, line 49: Removed all Streaming related test code.

src/main/java/com/mongodb/spark/sql/connector/read/MongoMicroBatchPartitionReader.java
```diff
@@ -0,0 +1,162 @@
...
+  }
+
+  private MongoChangeStreamCursor<BsonDocument> getCursor() {
+    if (mongoClient == null) {
+      mongoClient = readConfig.getMongoClient();
+    }
+    if (changeStreamCursor == null) {
+
+      List<BsonDocument> pipeline = new ArrayList<>();
+      pipeline.add(
```
rozza, line 118: The end bounds for this partition

src/main/java/com/mongodb/spark/sql/connector/read/MongoMicroBatchStream.java
```diff
@@ -0,0 +1,111 @@
...
+  }
+
+  @Override
+  public PartitionReaderFactory createReaderFactory() {
+    return new MongoMicroBatchPartitionReaderFactory(bsonDocumentToRowConverter, readConfig);
+  }
+
+  @Override
+  public Offset initialOffset() {
+    return new LongOffset(Instant.now().getEpochSecond());
```
rozza, line 94: Start from "now"

src/main/java/com/mongodb/spark/sql/connector/read/MongoMicroBatchPartitionReader.java
```diff
@@ -0,0 +1,162 @@
...
+      }
+      pipeline.addAll(partition.getPipeline());
+
+      ChangeStreamIterable<Document> changeStreamIterable =
+          mongoClient
+              .getDatabase(readConfig.getDatabaseName())
+              .getCollection(readConfig.getCollectionName())
+              .watch(pipeline)
+              .fullDocument(readConfig.getStreamFullDocument())
+              .startAtOperationTime(partition.getStartAtOperationTime());
```
jyemin, line 133: As discussed, let's consider whether this should start at 0, or unspecified (the latter is closer to what we do for continuous stream).

src/main/java/com/mongodb/spark/sql/connector/read/MongoMicroBatchPartitionReader.java
```diff
@@ -0,0 +1,162 @@
...
+
+  private MongoChangeStreamCursor<BsonDocument> getCursor() {
+    if (mongoClient == null) {
+      mongoClient = readConfig.getMongoClient();
+    }
+    if (changeStreamCursor == null) {
+
+      List<BsonDocument> pipeline = new ArrayList<>();
+      pipeline.add(
+          Aggregates.match(Filters.lt("clusterTime", partition.getEndOffsetTimestamp()))
```
jyemin, line 119: Confirm that 

- clusterTime corresponds to operationTime
- no events will be missed or duplicated

jyemin: Just added this as a note to self, as I don't fully understand that this is correct.

src/main/java/com/mongodb/spark/sql/connector/read/MongoMicroBatchPartitionReader.java
```diff
@@ -0,0 +1,162 @@
...
+        partition.getPartitionId(),
+        bsonDocumentToRowConverter.getSchema());
+  }
+
+  /** Proceed to next record, returns false if there is no more records. */
+  @Override
+  public boolean next() {
+    Assertions.ensureState(() -> !closed, () -> "Cannot call next() on a closed PartitionReader.");
+    MongoChangeStreamCursor<BsonDocument> cursor = getCursor();
+    BsonDocument next = cursor.tryNext();
```
jyemin, line 86: It's not correct to call `tryNext` here, I don't think. If `tryNext` returns null, it doesn't mean that it won't return non-null in the future.  I think what we need is to:

1. Remove the `resumeToken` filter from the pipeline
2. Replace `tryNext` with `next`
3. Return false from this method if `next.clusterTime` >= `endOffsetTimestamp`.

jyemin (changes requested):

src/main/java/com/mongodb/spark/sql/connector/read/MongoMicroBatchPartitionReader.java
```diff
@@ -0,0 +1,162 @@
...
+import com.mongodb.spark.sql.connector.config.ReadConfig;
+import com.mongodb.spark.sql.connector.exceptions.MongoSparkException;
+import com.mongodb.spark.sql.connector.schema.BsonDocumentToRowConverter;
+
+/**
+ * A partition reader returned by {@link
+ * MongoPartitionReaderFactory#createReader(org.apache.spark.sql.connector.read.InputPartition)}.
+ * It's responsible for outputting data for a RDD partition.
+ */
+public class MongoMicroBatchPartitionReader implements PartitionReader<InternalRow> {
```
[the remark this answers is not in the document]
rozza: No a single worker task will run this code.

src/main/java/com/mongodb/spark/sql/connector/read/MongoMicroBatchPartitionReader.java
```diff
@@ -0,0 +1,162 @@
...
+
+  private void releaseCursor() {
+    if (changeStreamCursor != null) {
+      LOGGER.debug("Closing cursor for partitionId: {}", partition.getPartitionId());
+      try {
+        changeStreamCursor.close();
+      } finally {
+        changeStreamCursor = null;
+        try {
+          mongoClient.close();
```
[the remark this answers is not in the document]
rozza: We own the MongoClient as we called ReadConfig.getMongoClient().  In the background it uses the MongoClientCache which returns an instance of CachedMongoClient, that handles the delayed closure of MongoClients so that they can be reused if multiple tasks on the same worker need it.

src/main/java/com/mongodb/spark/sql/connector/read/MongoScan.java
```diff
@@ -62,6 +63,20 @@ public Batch toBatch() {
...
+   * Returns the physical representation of this scan for streaming query with micro mode.
+   *
+   * <p>Utilizes MongoDBs change stream functionality, the continuous streams will consist of <a
+   * href="https://docs.mongodb.com/manual/reference/change-events/">change events</a>.
+   *
+   * <p>Note: Requires MongoDB 4.2+ To support continuing a change stream after a collection has
+   * been dropped.
+   */
+  @Override
+  public MicroBatchStream toMicroBatchStream(final String checkpointLocation) {
```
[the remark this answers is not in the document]
rozza: Added SPARK-379 to look into adding support.

src/main/java/com/mongodb/spark/sql/connector/read/MongoMicroBatchPartitionReader.java
```diff
@@ -0,0 +1,162 @@
...
+  public void close() {
+    if (!closed) {
+      closed = true;
+      releaseCursor();
+    }
+  }
+
+  private MongoChangeStreamCursor<BsonDocument> getCursor() {
+    if (mongoClient == null) {
+      mongoClient = readConfig.getMongoClient();
```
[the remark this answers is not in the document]
rozza: Done

src/main/java/com/mongodb/spark/sql/connector/read/MongoMicroBatchPartitionReader.java
rozza: Added SPARK-378 to track this.

Now defaults to -1 and will only set startAtOperationTime if >= 0. So eventually users will be able to chose.

src/main/java/com/mongodb/spark/sql/connector/read/MongoMicroBatchPartitionReader.java
rozza: Checked in #change-streams
> the clusterTime field of an event is exactly the same as the timestamp used in that event’s resume token

src/main/java/com/mongodb/spark/sql/connector/read/MongoMicroBatchPartitionReader.java
jyemin: > we are looking in the past so there should always data available now

But the server doesn't guarantee that it will send that data in the current batch.  It reserves the right to send empty batches at any time.  One realistic example: a write with operation time `x` where `x  < endOffsetTimestamp` but doesn't get majority committed until `y` where y` > endOffsetTimestamp`.

jyemin: > What I want to do avoid is any unnecessary blocking
 
I don't think we can avoid blocking given the micro-batch iterator-style API (it doesn't have a tryNext equivalent).

rozza: I think `tryNext` semantics could be used correctly here.  

* `tryNext()` will call `getMore()` as required. 
* It will return `null` if, the cursor no longer exists, the limit is reached or if there is no results from `getMore`.

We set no limit, so we just care about two scenarios. Cursor no longer exists (i.e closed) and no results available.

We can confirm that no results are ever expected by checking the postBatchResumeToken to ensure that if its timestamp is >= endOffsetTimestamp then there will be no future results.

That would stop any forever blocking which `next()` does if there are no matching results.

jyemin: I think cursor no longer exists results in an error returned from getMore, and is pretty rare (e.g. the collection has been dropped).  No results available is the tricky one.  As I said above, there is no guarantee from the server (afaik at least) that any getMore command response will return a result.  Empty batches are allowed, and are no guarantee that the next batch won't be empty.  So we can use tryNext here, but a loop would be required.  And isn't that effectively the same loop that the driver already has in its implementation on `next`?

src/main/java/com/mongodb/spark/sql/connector/read/MongoMicroBatchPartitionReader.java
```diff
@@ -0,0 +1,162 @@
...
+    Assertions.ensureState(() -> !closed, () -> "Cannot call next() on a closed PartitionReader.");
+    MongoChangeStreamCursor<BsonDocument> cursor = getCursor();
+    BsonDocument next = cursor.tryNext();
+    if (next != null) {
+      if (readConfig.streamPublishFullDocumentOnly()) {
+        next = next.getDocument(FULL_DOCUMENT, new BsonDocument());
+      }
+      currentRow = bsonDocumentToRowConverter.toInternalRow(next);
+    }
+    return next != null;
```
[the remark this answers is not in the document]
rozza: Ack - I've rewritten `next()` it now includes a loop until there is a result or the resume token is past the end offset.

src/main/java/com/mongodb/spark/sql/connector/read/MongoMicroBatchPartitionReader.java
rozza: This has been rewritten and should be clearer in intent now.

src/main/java/com/mongodb/spark/sql/connector/read/MongoMicroBatchPartitionReader.java
```diff
@@ -0,0 +1,191 @@
...
+   */
+  @Override
+  public boolean next() {
+    Assertions.ensureState(() -> !closed, () -> "Cannot call next() on a closed PartitionReader.");
+
+    MongoChangeStreamCursor<BsonDocument> cursor = getCursor();
+    BsonDocument cursorNext;
+
+    do {
+      cursorNext = cursor.tryNext();
```
jyemin, line 105: Do we need to wrap any `MongoException` thrown by `tryNext`, as we do in the `getCursor` method below?

jyemin (approved):
One question, but overall looks good.

src/main/java/com/mongodb/spark/sql/connector/read/MongoMicroBatchPartitionReader.java
rozza: Ack - sound good.

rozza:
Manually merging due to wrapping tryNext and the merge conflict in MongoScan

PR #86, opened 2022-11-18
InferSchema empty Array handling

rozza:
Ensures that empty arrays, don't force the loss of other documents array types when inferring the schema.

SPARK-375


jyemin (approved):
LGTM

Does this now make the behavior the same for 

* top-level document
* nested document in document
* nested document in array

?

rozza:
Thats correct :+1:

PR #90, opened 2023-02-02
Ensure partitioners can handle lack of count in storagestats

rozza:
Tests against timeseries where there is no count.

SPARK-387


src/integrationTest/java/com/mongodb/spark/sql/connector/read/MongoBatchTest.java
```diff
@@ -463,6 +469,66 @@ void testCustomMongoClientFactory() {
     assertTrue(CustomMongoClientFactory.CALLED.get());
   }
 
+  @Test
+  void testReadFromTimeseriesDatabase() {
+    assumeTrue(isAtLeastFiveDotZero() && !isSharded());
```
katcharov, line 474: What is the relationship between sharded and partitioners? That is, why is sharded excluded? (Just curious.)

katcharov (approved):
LGTM

src/integrationTest/java/com/mongodb/spark/sql/connector/read/MongoBatchTest.java
rozza: No relationship, its just that timeseries sharded support was added in 5.1 and that was causing issues.

PR #91, opened 2023-03-14
Allow to filter out null values from the document saved to the database

goshaQ:
Hello,

One of PRs recently merged to the main branch (#89) changed how the document written to the database looks like.
If previously fields containing null values were excluded, now they are stored in the database. And it is up to the user to filter out the fields containing null values. 

I see some issues with this approach:

1. For large dataframes containing many nullable columns there introduced significant overhead. Now each field has to be in the document written to the database and there will be many empty fields;
2. It is not really convinient to filter out the fields containing null values.

There is MongoDB Scala Driver which provides Macros for conversion of Scala case classes to documents written to the database. And depending on whether you want to keep null values or not it can be achieved with helpers `createCodecProvider` and `createCodecProviderIgnoreNone`. It would be great if this option could also be made available in this library.

To address this issue, I propose adding a new parameter called `ignoreNone` to the write configuration. If it is set to false (default value), the documents saved to the database will contain all the fields present in the dataframe. However, if it is set to true, it will filter out such values during conversion.

Note that for arrays (as well as map structures), any null values inside will also be filtered out.

This PR related to [SPARK-394](https://jira.mongodb.org/projects/SPARK/issues/SPARK-394).

### DIFF
--- a/src/main/java/com/mongodb/spark/sql/connector/config/WriteConfig.java
+++ b/src/main/java/com/mongodb/spark/sql/connector/config/WriteConfig.java
@@ -179,6 +179,21 @@ public final class WriteConfig extends AbstractMongoConfig {
 
   private static final boolean CONVERT_JSON_DEFAULT = false;
 
+  /**
+   * Keep null values in the BSON document.
+   *
+   * <p>Configuration: {@value}
+   *
+   * <p>Default: {@value IGNORE_NONE_DEFAULT}
+   *
+   * <p>If true, will store a field in the database only if it is present.
+   *
+   * @since 10.1
+   */
+  public static final String IGNORE_NONE_CONFIG = "ignoreNone";
+
+  private static final boolean IGNORE_NONE_DEFAULT = false;
+
   private final WriteConcern writeConcern;
   private final OperationType operationType;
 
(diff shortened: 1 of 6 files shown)


goshaQ:
@rozza 

Hello, would it be possible to have feedback on this PR? Should we expect such change to be introduced with upcoming release?

jhchee:
@goshaQ Thanks for contributing this! 
Anyone from the mongo team can take a look of this?

rozza:
Hi @goshaQ 

Thanks for the PR and apologies its taken the team a long time to get to this. I'll review for the next release.

cc @jhchee

src/main/java/com/mongodb/spark/sql/connector/config/WriteConfig.java
```diff
@@ -179,6 +179,21 @@ public String toString() {
...
+   *
+   * <p>Configuration: {@value}
+   *
+   * <p>Default: {@value IGNORE_NONE_DEFAULT}
+   *
+   * <p>If true, will store a field in the database only if it is present.
+   *
+   * @since 10.1
+   */
+  public static final String IGNORE_NONE_CONFIG = "ignoreNone";
```
rozza, line 193: nit: As this is a java code base and `None` is a scala type, I'll update to use `null` instead.

rozza (approved):
Thanks @goshaQ I like the PR.

I'll co-author and update the naming to use `null` just to be more explicit for any future maintainers of the codebase.

PR #93, opened 2023-03-31
Support escaped fieldnames in filters

rozza:
Spark automatically escapes some field names
eg ones containing a hyphen. Check and unescape any escaped field names before applying the filter
to MongoDB.

SPARK-393

### DIFF
--- a/src/main/java/com/mongodb/spark/sql/connector/read/MongoScanBuilder.java
+++ b/src/main/java/com/mongodb/spark/sql/connector/read/MongoScanBuilder.java
@@ -195,37 +195,42 @@ public final class MongoScanBuilder
       }
     } else if (filter instanceof EqualNullSafe) {
       EqualNullSafe equalNullSafe = (EqualNullSafe) filter;
+      String fieldName = getFieldName(equalNullSafe.attribute());
       return new FilterAndPipelineStage(
           filter,
-          getBsonValue(equalNullSafe.attribute(), equalNullSafe.value())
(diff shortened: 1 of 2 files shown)


jyemin (commented):
I think I'm missing something.  The description mentions hyphen characters, but I was only able to spot special handling of backticks.  Is the idea that hyphenated strings are surrounded by backticks?

jyemin (approved):
Assuming the answer to my question is "yes", then LGTM.

src/main/java/com/mongodb/spark/sql/connector/read/MongoScanBuilder.java
```diff
@@ -272,23 +281,30 @@ private FilterAndPipelineStage processFilter(final Filter filter) {
...
       return new FilterAndPipelineStage(
-          filter,
-          Filters.regex(stringStartsWith.attribute(), format("^%s.*", stringStartsWith.value())));
+          filter, Filters.regex(fieldName, format("^%s.*", stringStartsWith.value())));
     }
     return new FilterAndPipelineStage(filter, null);
   }
 
+  private String getFieldName(final String attribute) {
+    if (attribute.startsWith("`") && attribute.endsWith("`")) {
```
katcharov, line 302:
```suggestion
    // Spark automatically escapes hyphenated names using backticks
    if (attribute.startsWith("`") && attribute.endsWith("`")) {
```
Could also rename: 
`toJavaFieldName(final String sparkAttribute)`

katcharov (approved):
LGTM, just an optional comment

PR #95, opened 2023-06-06
IGNORE_NULL_VALUES Config

rozza:
SPARK-394

See: #91


rozza:
The only difference with the PR and how the `toJSON()` handler works is that `df.toJSON()` will not remove `null` values from inside arrays:

Take the following document / json
```
{'_id': 3, 'string': null, 'array': [null], 'subDoc': {'k':  'v'}}
```

df.toJSON() produces: 
```
{'_id': 3, 'array': [null], 'subDoc': {'k':  'v'}}
```
This PR with ignoreNullValues=true produces:
```
{'_id': 3, 'array': [], 'subDoc': {'k':  'v'}}

```

I think thats acceptable for the case of data storage but wanted to double check.

Ross

goshaQ (approved):
Everything looks good to me

PR #96, opened 2023-06-09
Multiple cursors for read stream microbatch

erinaldidb:
Improved readStream microbatch parallelism with multiple cursor generation

### DIFF
--- a/src/main/java/com/mongodb/spark/sql/connector/read/MongoMicroBatchPartitionReader.java
+++ b/src/main/java/com/mongodb/spark/sql/connector/read/MongoMicroBatchPartitionReader.java
@@ -167,6 +167,11 @@ final class MongoMicroBatchPartitionReader implements PartitionReader<InternalRo
       pipeline.add(
           Aggregates.match(Filters.lt("clusterTime", partition.getEndOffsetTimestamp()))
               .toBsonDocument());
+      if (partition.getStartOffsetTimestamp().getValue() > 0) {
+        pipeline.add(
+            Aggregates.match(Filters.gte("clusterTime", partition.getStartOffsetTimestamp()))
+                .toBsonDocument());
+      }
       if (readConfig.streamPublishFullDocumentOnly()) {
         pipeline.add(Aggregates.match(Filters.exists(FULL_DOCUMENT)).toBsonDocument());
       }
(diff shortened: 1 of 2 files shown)


rozza:
Hi @erinaldidb,

Thank you for this PR, I've added [SPARK-407](https://jira.mongodb.org/browse/SPARK-407) in our Jira to track this.

I fully understand the desire for parallelism when using Spark. In order to support Sparks micro batching mechanism the Spark connector uses time windows for batching and only produces a single micro batch partition per time window. The downside of this is it limits the parallelism - as only a single Spark worker can process a single micro batch at a time.  

This PR looks to increase the parallelism by producing multiple micro batches per time window. This would allow multiple spark workers to process partitions in parallel. However, this comes at a cost of processing data out of strict order as Spark does not guarantee processing partitions in order.

If multiple events for a document occur within the batch time window and multiple partitions receive an event for that document. The the order of processing the change events are not determined. 

Depending on the Spark processing logic this could easily cause subtle data corruption. 

The only safe scenario for this partitioning strategy would be if the collection was used as an insert only manner. Given this limititation I'm not sure how beneficial it would be for users.

If you have any further feedback or strategies please let me know,

Ross

erinaldidb:
Hi @rozza, thank you for investigating this PR.

I agree with you, this implementation will break processing ordering but by setting the parameter partitions to 1 (can be used as the default?) the behavior should be unchanged from the as-is. So this will give an option to the user to decide if they want to have ordering capabilities or performance improvement.

Also if there is too much data in one single partition the offset will increase every time and at some point, there will be too much data to be processed by one task and the spark job will fail.

Usually in a medallion architecture commonly used in databricks lakehouse the first layer, bronze, will have always append-only capability. Then next layer can take care of ordering and enrichments.

rozza:
Hi @erinaldidb,

BsonTimestamps are a special formatted type of timestamp, containing both a time portion and an increment portion. See: [Bson Timestmaps](https://www.mongodb.com/docs/manual/reference/bson-types/#timestamps)

> BSON has a special timestamp type for internal MongoDB use and is not associated with the regular [Date](https://www.mongodb.com/docs/manual/reference/bson-types/#std-label-document-bson-type-date) type. This internal timestamp type is a 64 bit value where:
>
> * the most significant 32 bits are a time_t value (seconds since the Unix epoch)
> * the least significant 32 bits are an incrementing ordinal for operations within a given second.

Because of this dividing the long value of a `BsonTimestamp` isn't a correct strategy and would lead to negative increment parts.

However, there are times when adding parallelism to change streams is acceptable but with the caveat that processing may not occur in as happens order. Given the makeup of a `BsonTimestamp` I think the lowest granularity should be a second. Attempting to go any lower by including ranges for the increment part of the timestamp risks producing empty partitions.

Please see #102 for that strategy

PR #97, opened 2023-06-13
Json string conversion improvements

rozza:
Added new config: `convertJson.nestedValuesOnly`
Which will only attempt to convert strings that contain nested Json Objects or Arrays.

Useful when dealing with extended Json and ensures strings that contain json like values are not automatically converted.

Example:

```
convertJson=true
convertJson.nestedValuesOnly=false

"{a: 1}"    => {a: 1}
"[1, 2, 3]" => [1, 2, 3]
"true"      => true
"0122345"   => 12345
"{a:b:c}"   => "{a:b:c}"

convertJson=true
convertJson.nestedValuesOnly=true

"{a: 1}"    => {a: 1}
"[1, 2, 3]" => [1, 2, 3]
"true"      => "true"
"0122345"   => "012345"
"{a:b:c}"   => "{a:b:c}"
```

SPARK-388 SPARK-405

### DIFF
--- a/src/main/java/com/mongodb/spark/sql/connector/config/WriteConfig.java
+++ b/src/main/java/com/mongodb/spark/sql/connector/config/WriteConfig.java
@@ -88,16 +88,15 @@ public final class WriteConfig extends AbstractMongoConfig {
     }
 
     static ConvertJson fromString(final String jsonType) {
-      for (ConvertJson convertJsonType : ConvertJson.values()) {
-        if (jsonType.equalsIgnoreCase(TRUE)) {
-          LOGGER.warn("{}: 'true' is deprecated. Use: '{}' instead.", CONVERT_JSON_CONFIG, ANY);
-          return ANY;
-        }
-        if (jsonType.equalsIgnoreCase(convertJsonType.value)) {
-          return convertJsonType;
-        }
+      if (jsonType.equalsIgnoreCase(TRUE)) {
+        LOGGER.warn("{}: '{}' is deprecated. Use: '{}' instead.", CONVERT_JSON_CONFIG, TRUE, ANY);
+        return ANY;
       }
-      throw new ConfigException(format("'%s' is not a valid Convert Json Type", jsonType));
+      try {
+        return ConvertJson.valueOf(jsonType.toUpperCase(Locale.ROOT));
+      } catch (IllegalArgumentException e) {
+        throw new ConfigException(format("'%s' is not a valid Convert Json Type", jsonType), e);
+      }
     }
 
     @Override
--- a/src/main/java/com/mongodb/spark/sql/connector/config/WriteConfig.java
+++ b/src/main/java/com/mongodb/spark/sql/connector/config/WriteConfig.java
@@ -22,6 +22,7 @@ import static java.util.Collections.singletonList;
 
 import java.util.HashMap;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 import java.util.concurrent.TimeUnit;
 
@@ -77,7 +78,7 @@ public final class WriteConfig extends AbstractMongoConfig {
     FALSE("false"),
     /** Try to parse any string as a json value */
     ANY("any"),
-    /** Try to parse any string as a json value */
+    /** Only try to parse strings are potentially json objects or arrays */
     OBJECT_OR_ARRAY_ONLY("objectOrArrayOnly");
 
     private final String value;
--- a/src/main/java/com/mongodb/spark/sql/connector/config/WriteConfig.java
+++ b/src/main/java/com/mongodb/spark/sql/connector/config/WriteConfig.java
@@ -25,6 +25,9 @@ import java.util.List;
 import java.util.Map;
 import java.util.concurrent.TimeUnit;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import com.mongodb.WriteConcern;
 
 import com.mongodb.spark.sql.connector.exceptions.ConfigException;
@@ -36,6 +39,7 @@ import com.mongodb.spark.sql.connector.exceptions.ConfigException;
  */
 public final class WriteConfig extends AbstractMongoConfig {
   private static final long serialVersionUID = 1L;
+  private static final Logger LOGGER = LoggerFactory.getLogger(WriteConfig.class);
 
   /** The operation type for the write. */
   public enum OperationType {
@@ -59,6 +63,41 @@ public final class WriteConfig extends AbstractMongoConfig {
         }
       }
       throw new ConfigException(format("'%s' is not a valid Write Operation Type", operationType));
+    }
+
+    @Override
+    public String toString() {
+      return value;
+    }
+  }
+
+  /** The convert json configuration. */
+  public enum ConvertJson {
+    /** No conversion of string types */
+    FALSE("false"),
+    /** Try to parse any string as a json value */
+    ANY("any"),
+    /** Try to parse any string as a json value */
+    OBJECT_OR_ARRAY_ONLY("objectOrArrayOnly");
+
+    private final String value;
+    private static final String TRUE = "true";
+
+    ConvertJson(final String operationType) {
+      this.value = operationType;
+    }
+
+    static ConvertJson fromString(final String jsonType) {
+      for (ConvertJson convertJsonType : ConvertJson.values()) {
+        if (jsonType.equalsIgnoreCase(TRUE)) {
+          LOGGER.warn("{}: 'true' is deprecated. Use: '{}' instead.", CONVERT_JSON_CONFIG, ANY);
+          return ANY;
+        }
+        if (jsonType.equalsIgnoreCase(convertJsonType.value)) {
+          return convertJsonType;
+        }
+      }
+      throw new ConfigException(format("'%s' is not a valid Convert Json Type", jsonType));
     }
 
     @Override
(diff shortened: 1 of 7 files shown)


stIncMale:
The example `"0122345"   => "12345"` for
```
convertJson=true
convertJson.nestedValuesOnly=true
```

should be `"0122345"   => "012345"`

stIncMale:
Let's change the type of the existing `spark.mongodb.write.convertJson` property from a Boolean to enumeration (I don't know if our Spark connector properties have types enforced by validation, but at least conceptually, we may thing of them as having a type) with the following values:

- `false` - Default. Is equivalent to not specifying the property.
- `true` - Always attempt to convert a [Spark string](https://spark.apache.org/docs/3.1.2/sql-ref-datatypes.html) value into a [BSON](https://www.mongodb.com/docs/manual/reference/bson-types/) value by treating the string value as either the [canonical or relaxed extended JSON](https://www.mongodb.com/docs/manual/reference/mongodb-extended-json/) value.
- `nestedValuesOnly` - Similar to `true`, but attempts are made only if the Spark string value looks like a [JSON object](https://www.rfc-editor.org/rfc/rfc8259.html#section-4) or [JSON array](https://www.rfc-editor.org/rfc/rfc8259.html#section-5), i.e., is encased between either `{` and `}`, or `[` and `]`.

P.S. Consider renaming `nestedValuesOnly` to `objectOrArrayOnly`: the behavior is to attempt to convert all Spark strings that look like a JSON object or array, even if they are top-level object/array, and even if they don't have any nested objects/arrays.

stIncMale (commented):
Let's change the type of the existing `spark.mongodb.write.convertJson` property from Boolean to enumeration (I don't know if our Spark connector properties have types enforced by validation, but at least conceptually, we may think of them as having a type) with the following values:

- `false` - Default. Is equivalent to not specifying the property.
- `true` - Always attempt to convert a [Spark string](https://spark.apache.org/docs/latest/sql-ref-datatypes.html) value into a [BSON](https://www.mongodb.com/docs/manual/reference/bson-types/) value by treating the string value as either the [canonical or relaxed extended JSON](https://www.mongodb.com/docs/manual/reference/mongodb-extended-json/) value.
- `nestedValuesOnly` - Similar to `true`, but attempts are made only if the Spark string value looks like a [JSON object](https://www.rfc-editor.org/rfc/rfc8259.html#section-4) or [JSON array](https://www.rfc-editor.org/rfc/rfc8259.html#section-5) value, i.e., is encased between either `{` and `}`, or `[` and `]`.

P.S. Consider renaming `nestedValuesOnly` to `objectOrArrayOnly`: the behavior is to attempt to convert all Spark strings that look like a JSON object or array, even if they are top-level object/array, and even if they don't have any nested objects/arrays.

rozza:
Updated to use a single configuration / enum.

As SPARK-406 has been backlogged due to no clear user demand at this time, I opted for the simplest recommended approach.

src/main/java/com/mongodb/spark/sql/connector/config/WriteConfig.java
```diff
@@ -67,6 +71,41 @@ public String toString() {
...
+    private static final String TRUE = "true";
+
+    ConvertJson(final String operationType) {
+      this.value = operationType;
+    }
+
+    static ConvertJson fromString(final String jsonType) {
+      for (ConvertJson convertJsonType : ConvertJson.values()) {
+        if (jsonType.equalsIgnoreCase(TRUE)) {
+          LOGGER.warn("{}: 'true' is deprecated. Use: '{}' instead.", CONVERT_JSON_CONFIG, ANY);
```
rozza, line 93: I didn't feel the need to introduce a TRUE - value rather opted just to log that its deprecated.

src/main/java/com/mongodb/spark/sql/connector/config/WriteConfig.java
```diff
@@ -67,6 +71,41 @@ public String toString() {
...
+          return ANY;
+        }
+        if (jsonType.equalsIgnoreCase(convertJsonType.value)) {
+          return convertJsonType;
+        }
+      }
+      throw new ConfigException(format("'%s' is not a valid Convert Json Type", jsonType));
+    }
```
stIncMale, line 101: - There is no need to check `if (jsonType.equalsIgnoreCase(TRUE))` in a loop.
- Instead of looping, we can use the existing `ConvertJson.valueOf` method.


```suggestion
    static ConvertJson fromString(final String jsonType) {
      if (jsonType.equalsIgnoreCase(TRUE)) {
        LOGGER.warn("{}: '{}' is deprecated. Use: '{}' instead.", CONVERT_JSON_CONFIG, TRUE, ANY);
        return ANY;
      }
      try {
        return ConvertJson.valueOf(jsonType.toUpperCase(Locale.ROOT));
      } catch (IllegalArgumentException e) {
        throw new ConfigException(format("'%s' is not a valid Convert Json Type", jsonType), e);
      }
    }
```

src/test/java/com/mongodb/spark/sql/connector/schema/SchemaTest.java
```diff
@@ -342,6 +342,43 @@ abstract class SchemaTest {
...
+  static final BsonDocument CONVERT_JSON_DOCUMENT =
+      BsonDocument.parse(
+          "{"
+              + " \"booleanString\": true,"
+              + " \"numericString\": 012345,"
+              + " \"document\": {\"a\": \"foo\", \"b\": \"012345\", \"c\": \"true\"},"
+              + " \"array\": [\"012345\", \"true\", {\"a\": \"foo\", \"b\": \"012345\", \"c\": \"true\"}]"
+              + "}");
+
+  static final BsonDocument CONVERT_JSON_NESTED_ONLY_DOCUMENT =
```
stIncMale, line 373:
```suggestion
  static final BsonDocument CONVERT_JSON_OBJECT_OR_ARRAY_ONLY_DOCUMENT =
```

src/test/java/com/mongodb/spark/sql/connector/schema/SchemaTest.java
```diff
@@ -342,6 +342,43 @@ abstract class SchemaTest {
               .toArray(),
           BSON_DOCUMENT_STRING_SCHEMA);
 
+  // JSON Conversion types handling
+  static final StructType CONVERT_JSON_SCHEMA =
```
stIncMale, line 346:
```suggestion
  private static final StructType CONVERT_JSON_SCHEMA =
```

src/test/java/com/mongodb/spark/sql/connector/schema/RowToBsonDocumentConverterTest.java
```diff
@@ -42,18 +42,24 @@
...
 public class RowToBsonDocumentConverterTest extends SchemaTest {
   private static final RowToBsonDocumentConverter DEFAULT_CONVERTER =
-      new RowToBsonDocumentConverter(new StructType(), false, false);
-  private static final RowToBsonDocumentConverter EXTENDED_JSON_CONVERTER =
-      new RowToBsonDocumentConverter(new StructType(), true, false);
+      new RowToBsonDocumentConverter(new StructType(), WriteConfig.ConvertJson.FALSE, false);
+  private static final RowToBsonDocumentConverter JSON_CONVERTER =
+      new RowToBsonDocumentConverter(new StructType(), WriteConfig.ConvertJson.ANY, false);
+
+  private static final RowToBsonDocumentConverter NESTED_JSON_CONVERTER =
```
stIncMale, line 57:
```suggestion
  private static final RowToBsonDocumentConverter OBJECT_OR_ARRAY_JSON_CONVERTER =
```

src/test/java/com/mongodb/spark/sql/connector/schema/RowToBsonDocumentConverterTest.java
```diff
@@ -62,17 +68,18 @@ void testSimpleTypes() {
...
+  @DisplayName("test json converter all fields")
+  void testJsonConverter() {
+    assertEquals(BSON_DOCUMENT_RELAXED, JSON_CONVERTER.fromRow(ALL_TYPES_RELAXED_JSON_ROW));
+    assertEquals(BSON_DOCUMENT_ALL_TYPES, JSON_CONVERTER.fromRow(ALL_TYPES_EXTENDED_JSON_ROW));
+    assertEquals(CONVERT_JSON_DOCUMENT, JSON_CONVERTER.fromRow(CONVERT_JSON_ROW));
   }
 
   @Test
-  @DisplayName("test extended json string types")
+  @DisplayName("test json converter nested fields only")
```
stIncMale, line 79:
```suggestion
  @DisplayName("test json converter objects or arrays only")
```

src/main/java/com/mongodb/spark/sql/connector/schema/RowToBsonDocumentConverter.java
```diff
@@ -226,4 +241,17 @@ private BsonValue processString(final String data) {
...
+      case FALSE:
+        return false;
+      case ANY:
+        return true;
+      case OBJECT_OR_ARRAY_ONLY:
+        return isJsonObjectOrArray(data);
+      default:
+        throw new IllegalStateException("Unexpected value: " + convertJson);
+    }
+  }
```
stIncMale, line 256: `IllegalStateException` may only be thrown here if there is a bug in the MongoDB Spark Connector: we introduced a new `enum` constant of type `WriteConfig.ConvertJson`, but forgot to add a `case` for it here. In the MongoDB Java Driver we throw `AssertionError` such cases (see `com.mongodb.assertions.Assertions`) to make it clear that there is a driver bug. Do we want to start using the same approach in the MongoDB Spark Connector?

stIncMale (changes requested):

src/main/java/com/mongodb/spark/sql/connector/config/WriteConfig.java
```diff
@@ -67,6 +71,41 @@ public String toString() {
...
   }
 
+  /** The convert json configuration. */
+  public enum ConvertJson {
+    /** No conversion of string types */
+    FALSE("false"),
+    /** Try to parse any string as a json value */
+    ANY("any"),
+    /** Try to parse any string as a json value */
+    OBJECT_OR_ARRAY_ONLY("objectOrArrayOnly");
```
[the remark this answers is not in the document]
rozza: Done

src/main/java/com/mongodb/spark/sql/connector/schema/RowToBsonDocumentConverter.java
rozza: Done

stIncMale (approved):

PR #98, opened 2023-06-21
Added stream resumability

rozza:
Adds a MongoOffsetStore for storing and retrieving MongoOffsets.

Converted both the micro batch and the continuous streams to read from and write to the MongoOffsetStore.

SPARK-379


rozza:
Going to merge with startAtOperationTime as well.

PR #102, opened 2023-07-03
Added `change.stream.micro.batch.max.partition.count` config

rozza:
Support partitioning of micro batches up to a max
number of partitions.

SPARK-407

### DIFF
--- a/src/main/java/com/mongodb/spark/sql/connector/read/MongoInputPartitionHelper.java
+++ b/src/main/java/com/mongodb/spark/sql/connector/read/MongoInputPartitionHelper.java
@@ -24,11 +24,13 @@ import java.util.Arrays;
 import java.util.List;
 import java.util.Optional;
 import java.util.function.Function;
+import java.util.stream.IntStream;
 
 import org.apache.spark.sql.types.StructType;
 
 import org.bson.BsonDocument;
 import org.bson.BsonInt32;
+import org.bson.BsonTimestamp;
 
 import com.mongodb.spark.sql.connector.config.ReadConfig;
 import com.mongodb.spark.sql.connector.exceptions.MongoSparkException;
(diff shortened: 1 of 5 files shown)


src/main/java/com/mongodb/spark/sql/connector/read/MongoInputPartitionHelper.java
```diff
@@ -78,6 +80,46 @@ static MongoInputPartition[] generateMongoBatchPartitions(
...
+        previous = next;
+      }
+    }
+    partitions.add(endTimestamp);
+
+    return IntStream.range(1, partitions.size())
+        .mapToObj(
+            i ->
+                new MongoMicroBatchInputPartition(
+                    i,
```
stIncMale, line 116: 1. Previously `MongoMicroBatchStream.planInputPartitions` used 0 as the minimal `MongoInputPartition.partitionId`, here 1 is used. Is that important?
2. Previously `MongoMicroBatchStream.planInputPartitions` remembered all used `partitionId` values (by remembering the largest one) and avoided reusing the same `partitionId` if called multiple times (its documentation states that _"will be called multiple times"_). With the changes in this PR, `partitionId` of partitions produced by multiple calls to `MongoMicroBatchStream.planInputPartitions` will intersect. Is this OK? If yes, what is the purpose of `MongoInputPartition.partitionId`?

rozza: Partition Id is only used for logging purposes and is used to determine if a cursor has been opened & subsequently closed.

I considered reusing an ever increasing Id - but as these partitions are batch based and all partitions have to be consumed together within that batch before the next batch is created, I didn't consider it necessary.

So in short, the purpose of the Id is limited but occasionally useful in debugging.

stIncMale (approved):

PR #105, opened 2023-11-07
Use $ne for isNotNull

rozza:
~[$ne](https://www.mongodb.com/docs/manual/reference/operator/query/ne/#mongodb-query-op.-ne) also excludes documents where the field doesn't exist (this used to not be the case).~

SPARK-410


jyemin:
> [$ne](https://www.mongodb.com/docs/manual/reference/operator/query/ne/#mongodb-query-op.-ne) also excludes documents where the field doesn't exist (this used to not be the case)

I'm curious when the behavior changed.  I can't find any documentation of that.  Do you know?

rozza:
@jyemin appears I'm mistaken. Seems $ne in the docs back to 2.6 says "This includes documents that do not contain the field."  

Perhaps misunderstanding that sentance, might have been the cause of my belief!

jyemin (approved):
OK so it will work the same for all server versions we support!

LGTM

PR #107, opened 2024-01-22
Stop calling `MongoMicroBatchPartitionReader.tryNext` if the resume token timestamp is equal to the end offset timestamp

stIncMale:
SPARK-416


rozza (approved):
LGTM!

PR #109, opened 2024-02-07
Support scanning from multiple collections

stIncMale:
For the description of the new functionality and configuration see the API documentation of the following program elements:

- `MongoConfig.COLLECTION_NAME_CONFIG`
- `CollectionsConfig.Type` and its values.

For the complete picture, please see also [this](https://jira.mongodb.org/browse/SPARK-404?focusedCommentId=6018950&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-6018950) comment.

Note that I marked all the new `public` program elements as `@ApiStatus.Internal`. I am not sure whether that is correct, but I don't see myself why they need to be part of the API, nor do I see why all the existing program elements that allow accessing the configuration for reading had to be made part of the API previously.

SPARK-404

### DIFF
--- a/src/integrationTest/java/com/mongodb/spark/sql/connector/mongodb/MongoSparkConnectorTestCase.java
+++ b/src/integrationTest/java/com/mongodb/spark/sql/connector/mongodb/MongoSparkConnectorTestCase.java
@@ -17,6 +17,7 @@ package com.mongodb.spark.sql.connector.mongodb;
 
 import static java.lang.String.format;
 import static java.util.Arrays.asList;
+import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.junit.jupiter.api.Assumptions.assumeFalse;
@@ -176,10 +177,19 @@ public class MongoSparkConnectorTestCase {
     MongoCollection<Document> profileCollection = database.getCollection("system.profile");
     try {
       profileCollection.drop();
-      database.runCommand(BsonDocument.parse(
-          format("{profile: 2, filter: {ns: '%s.%s'}}", databaseName, collectionName)));
+      database.runCommand(new Document("profile", 1)
+          .append(
+              "filter",
+              Filters.eq("ns", new MongoNamespace(databaseName, collectionName).getFullName())));
 
       runnable.run();
+
+      List<Document> profileDocsWithComment = profileCollection
+          .find(Filters.eq("command.comment", TEST_COMMENT))
+          .into(new ArrayList<>());
+      assertFalse(
+          profileDocsWithComment.isEmpty(),
+          format("There were no commands observed with the comment \"%s\"", TEST_COMMENT));
 
       List<Document> profileDocs = profileCollection
           .find(Filters.nor(
--- a/src/integrationTest/java/com/mongodb/spark/sql/connector/read/AbstractMongoStreamTest.java
+++ b/src/integrationTest/java/com/mongodb/spark/sql/connector/read/AbstractMongoStreamTest.java
@@ -732,7 +732,7 @@ abstract class AbstractMongoStreamTest extends MongoSparkConnectorTestCase {
       final BiConsumer<String, MongoCollection<BsonDocument>> biConsumer) {
     return mongoConfig -> {
       if (msg != null) {
-        LOGGER.info("-> With source: " + msg);
+        LOGGER.info("-> With source: {}", msg);
       }
       ReadConfig readConfig = mongoConfig.toReadConfig();
       MongoCollection<BsonDocument> collection = readConfig.withClient(client -> client
@@ -746,7 +746,7 @@ abstract class AbstractMongoStreamTest extends MongoSparkConnectorTestCase {
       @Nullable final String msg, final BiConsumer<String, MongoDatabase> biConsumer) {
     return mongoConfig -> {
       if (msg != null) {
-        LOGGER.info("-> With source: " + msg);
+        LOGGER.info("-> With source: {}", msg);
       }
       ReadConfig readConfig = mongoConfig.toReadConfig();
       MongoDatabase db =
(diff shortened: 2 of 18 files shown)


src/integrationTest/java/com/mongodb/spark/sql/connector/mongodb/MongoSparkConnectorTestCase.java
```diff
@@ -176,11 +177,20 @@ private void assertCommentsInProfile(
...
+              Filters.eq("ns", new MongoNamespace(databaseName, collectionName).getFullName())));
 
       runnable.run();
 
+      List<Document> profileDocsWithComment = profileCollection
+          .find(Filters.eq("command.comment", TEST_COMMENT))
+          .into(new ArrayList<>());
+      assertFalse(
+          profileDocsWithComment.isEmpty(),
+          format("There were no commands observed with the comment \"%s\"", TEST_COMMENT));
```
stIncMale, line 192: Here we profile less, and we assert that we still observe commands with the `TEST_COMMENT`. I.e., this change makes the test stricter, it makes it actually test what is supposed to be tested.

src/main/java/com/mongodb/spark/sql/connector/schema/InferSchema.java
```diff
@@ -65,22 +69,42 @@ public final class InferSchema {
...
+      samplePipeline.add(Aggregates.sample(readConfig.getInferSchemaSampleSize()));
+      for (String collectionName : collectionNames) {
+        database
+            .getCollection(collectionName, BsonDocument.class)
+            .aggregate(samplePipeline)
             .allowDiskUse(readConfig.getAggregationAllowDiskUse())
             .comment(readConfig.getComment())
-            .into(new ArrayList<>())),
-        readConfig);
+            .into(sampleDocuments);
```
stIncMale, line 104: We are sampling all the collections sequentially. If there are many collections, this may take significant amount of time. If users get back to us pointing that this is a problem, then we may think about possible solutions and implement them.

rozza (changes requested):
Only minor changes requested. 

However, I have serious doubts on the usefulness of inferring the schema across multiple collections. 

Given general database modelling of collections (tables) storing things that are alike. The chances of multiple tables containing similar data is reduced. I think with the flexibility of document databases, where collections can easily store similar datatypes this is even further reduced.  So realistically, because it can be done doesn't equate to it should be done. 

Do we have data that suggests this would be practical for users?

If this does go ahead we need to ensure better config validation / regression testing for batched reads. Currently it errors with:

```
Partitioning failed. The connector is configured to access MongoSparkConnectorTest.*, which is not supported in the current context.
```

But I think the error could be clearer and validated before the partitioner helper and the error message should mention that batch reads only support single collections.

src/main/java/com/mongodb/spark/sql/connector/schema/InferSchema.java
stIncMale: That was my original approach, but Prakul explicitly asked for the schema inference, and said that it is an important feature.

src/main/java/com/mongodb/spark/sql/connector/config/CollectionsConfig.java
```diff
@@ -0,0 +1,279 @@
...
+  @ApiStatus.Internal
+  public enum Type {
+    /**
+     * Scan a {@linkplain MongoConfig#getCollectionName() single collection}.
+     * This type is supported by
+     * {@linkplain Scan#toBatch() batch queries},
+     * {@linkplain Scan#toMicroBatchStream(String) micro-batch streams},
+     * {@linkplain Scan#toContinuousStream(String) continuous streams}.
+     */
+    @ApiStatus.Internal
```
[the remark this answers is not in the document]
stIncMale: Yes, but I wanted to reduce the chances of a user missing the annotation. If you think that's an overkill, I can remove the annotation here.

src/integrationTest/java/com/mongodb/spark/sql/connector/read/AbstractMongoStreamTest.java
```diff
@@ -576,18 +728,45 @@ private StreamingQuery createStreamingQuery(
...
+          .getCollection(collectionName(), BsonDocument.class));
+      biConsumer.accept(msg, collection);
+    };
+  }
+
+  private Consumer<MongoConfig> withSourceDb(
+      @Nullable final String msg, final BiConsumer<String, MongoDatabase> biConsumer) {
+    return mongoConfig -> {
+      if (msg != null) {
+        LOGGER.info("-> With source: " + msg);
```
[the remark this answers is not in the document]
stIncMale: I copied that code and didn't pay attention :( Done.

src/main/java/com/mongodb/spark/sql/connector/config/MongoConfig.java
```diff
@@ -151,8 +153,38 @@ static WriteConfig writeConfig(final Map<String, String> options) {
    */
   String DATABASE_NAME_CONFIG = "database";
 
+  // This documentation links to `WriteBuilder` instead of `Write`
+  // because `Write` was added in park-catalyst 3.2.0, and we must support 3.1.2.
```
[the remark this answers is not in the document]
stIncMale: Done.

stIncMale:
> I think the error could be clearer and validated before the partitioner helper

Done.

src/main/java/com/mongodb/spark/sql/connector/config/CollectionsConfig.java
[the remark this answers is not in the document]
rozza: Cool - it doesn't really matter as user won't use *any* of this API in their Spark App.  Just the Spark API which then calls down to the connector abstraction.

src/main/java/com/mongodb/spark/sql/connector/schema/InferSchema.java
rozza: Fair enough - not sold its going to be practical, but I suppose for those that need multiple collection streaming it could be 🤷

prakul: Slightly a noob question - can the user still add the schema definition explicitly (instead of auto-inference), if they chose to, when using multiple collections? 
We should be calling out in the documentation very clearly the suggestion to set schema explicitly, otherwise bear the performance implications. (We get this feedback consistently from user that they don't want to specify explicit schema as it tends to change over time, etc.)

stIncMale: Yes, a user can specify an explicit schema as described in https://www.mongodb.com/docs/spark-connector/current/streaming-mode/streaming-read/.

PR #110, opened 2024-02-29
Cache the DataType to BsonValue mapping

rozza:
Improves performance as its only done once rather than per row.

Returns performance to the 3.x branch as that also memoizes the data type to bson value mapping function for each field.

SPARK-419

### DIFF
--- a/src/main/java/com/mongodb/spark/sql/connector/read/MongoScanBuilder.java
+++ b/src/main/java/com/mongodb/spark/sql/connector/read/MongoScanBuilder.java
@@ -28,13 +28,13 @@ import com.mongodb.spark.sql.connector.assertions.Assertions;
 import com.mongodb.spark.sql.connector.config.MongoConfig;
 import com.mongodb.spark.sql.connector.config.ReadConfig;
 import com.mongodb.spark.sql.connector.config.WriteConfig;
+import com.mongodb.spark.sql.connector.schema.RowToBsonDocumentConverter;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Locale;
 import java.util.Optional;
 import java.util.Set;
-import java.util.function.Function;
 import java.util.stream.Collectors;
 import org.apache.spark.sql.SparkSession;
 import org.apache.spark.sql.connector.read.Scan;
@@ -305,7 +305,7 @@ public final class MongoScanBuilder
           localSchema = (StructType) localField.dataType();
         }
       }
-      Function<Object, BsonValue> objectToBsonValue =
+      RowToBsonDocumentConverter.ObjectToBsonValue objectToBsonValue =
           createObjectToBsonValue(localDataType, WriteConfig.ConvertJson.FALSE, false);
       return Optional.of(objectToBsonValue.apply(value));
     } catch (Exception e) {
--- a/src/main/java/com/mongodb/spark/sql/connector/read/MongoScanBuilder.java
+++ b/src/main/java/com/mongodb/spark/sql/connector/read/MongoScanBuilder.java
@@ -17,6 +17,7 @@
 
 package com.mongodb.spark.sql.connector.read;
 
+import static com.mongodb.spark.sql.connector.schema.RowToBsonDocumentConverter.createObjectToBsonValue;
 import static java.lang.String.format;
 import static java.util.Collections.emptyList;
 import static java.util.Collections.singletonList;
@@ -27,13 +28,13 @@ import com.mongodb.spark.sql.connector.assertions.Assertions;
 import com.mongodb.spark.sql.connector.config.MongoConfig;
 import com.mongodb.spark.sql.connector.config.ReadConfig;
 import com.mongodb.spark.sql.connector.config.WriteConfig;
-import com.mongodb.spark.sql.connector.schema.RowToBsonDocumentConverter;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Locale;
 import java.util.Optional;
 import java.util.Set;
+import java.util.function.Function;
 import java.util.stream.Collectors;
 import org.apache.spark.sql.SparkSession;
 import org.apache.spark.sql.connector.read.Scan;
@@ -69,8 +70,6 @@ import org.jetbrains.annotations.Nullable;
 @ApiStatus.Internal
 public final class MongoScanBuilder
     implements ScanBuilder, SupportsPushDownFilters, SupportsPushDownRequiredColumns {
-  private static final RowToBsonDocumentConverter CONVERTER =
-      new RowToBsonDocumentConverter(new StructType(), WriteConfig.ConvertJson.FALSE, false);
   private final StructType schema;
   private final ReadConfig readConfig;
   private final boolean isCaseSensitive;
@@ -306,7 +305,9 @@ public final class MongoScanBuilder
           localSchema = (StructType) localField.dataType();
         }
       }
-      return Optional.of(CONVERTER.toBsonValue(localDataType, value));
+      Function<Object, BsonValue> objectToBsonValue =
+          createObjectToBsonValue(localDataType, WriteConfig.ConvertJson.FALSE, false);
+      return Optional.of(objectToBsonValue.apply(value));
     } catch (Exception e) {
       // ignore
       return Optional.empty();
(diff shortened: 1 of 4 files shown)


src/main/java/com/mongodb/spark/sql/connector/schema/RowToBsonDocumentConverter.java
```diff
@@ -105,54 +113,86 @@ public BsonDocument fromRow(final InternalRow row) {
...
         BigDecimal bigDecimal = data instanceof BigDecimal
             ? (BigDecimal) data
             : ((Decimal) data).toBigDecimal().bigDecimal();
         return new BsonDecimal128(new Decimal128(bigDecimal));
-      } else if (dataType instanceof ArrayType) {
+      };
+    } else if (dataType instanceof ArrayType) {
+      return (data) -> {
         DataType elementType = ((ArrayType) dataType).elementType();
+        Function<Object, BsonValue> element =
```
jyemin, line 194: To avoid a performance penalty here, should this be lifted out of the lambda?

src/main/java/com/mongodb/spark/sql/connector/schema/RowToBsonDocumentConverter.java
```diff
@@ -165,17 +205,21 @@ public BsonValue toBsonValue(final DataType dataType, final Object data) {
...
+      };
+    } else if (dataType instanceof MapType) {
+      return (data) -> {
         DataType keyType = ((MapType) dataType).keyType();
         DataType valueType = ((MapType) dataType).valueType();
         if (!(keyType instanceof StringType)) {
           throw new DataException(
               format("Cannot cast %s into a BsonValue. Invalid key type %s.", data, keyType));
         }
+        Function<Object, BsonValue> value =
```
jyemin, line 221: To avoid a performance penalty here, should this be lifted out of the lambda?

jyemin (changes requested):

src/main/java/com/mongodb/spark/sql/connector/schema/RowToBsonDocumentConverter.java
```diff
@@ -248,4 +304,20 @@ private boolean parseJsonData(final String data) {
...
+          format(
+              "Unable to create InternalRow to Row Function using %s. Errors: %s",
+              schema, e.getMessage()),
+          e);
+    }
+  }
+
+  interface ObjectToBsonElement extends Function<Object, BsonElement>, Serializable {}
+
+  interface ObjectToBsonValue extends Function<Object, BsonValue>, Serializable {}
```
stIncMale, line 322: We don't seem to test directly whether the objects of types `ObjectToBsonElement`, `ObjectToBsonValue` that we create can actually be serialized. How sure are we that they are?

A quick glance suggests that the `ObjectToBsonElement` created in `dataTypeToBsonElement` may not be serializable in practice because it captures `field` of the `StructField` type, which does not seem to be `Serializable`. It also indirectly captures `ConvertJson`, which is also not `Serializable`.

stIncMale (changes requested):
@rozza Have you confirmed with manual experiments (as far as I know, we don't have anything automated) that the change has positive impact on the performance characteristics?

src/main/java/com/mongodb/spark/sql/connector/schema/RowToBsonDocumentConverter.java
```diff
@@ -185,30 +229,40 @@ public BsonValue toBsonValue(final DataType dataType, final Object data) {
...
+          "Cannot cast %s into a BsonValue. %s data type has no matching BsonValue.",
+          data, dataType));
+    };
+  }
+
+  private static BsonDocument rowToBsonDocument(
+      final Row row,
+      final List<ObjectToBsonElement> objectToBsonElements,
+      final boolean ignoreNulls) {
+    List<BsonElement> elements = new ArrayList<>();
```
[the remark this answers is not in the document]
rozza: I originally did and it had no meaningful impact on the code and speed. But can revert.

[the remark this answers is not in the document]
rozza: Done

src/main/java/com/mongodb/spark/sql/connector/schema/RowToBsonDocumentConverter.java
rozza: Done.

src/main/java/com/mongodb/spark/sql/connector/schema/RowToBsonDocumentConverter.java
rozza: Done

src/main/java/com/mongodb/spark/sql/connector/schema/RowToBsonDocumentConverter.java
```diff
@@ -105,54 +113,86 @@ public BsonDocument fromRow(final InternalRow row) {
...
-   * @return the bsonValue
+   * @param dataType the data type of the object
+   * @param convertJson the string json conversion setting
+   * @param ignoreNulls true if ignoring null values
+   * @return a serializable function converts an object to a BsonValue
    */
-  @SuppressWarnings("unchecked")
-  public BsonValue toBsonValue(final DataType dataType, final Object data) {
-    try {
+  public static ObjectToBsonValue createObjectToBsonValue(
```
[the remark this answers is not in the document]
rozza: Done

src/main/java/com/mongodb/spark/sql/connector/schema/RowToBsonDocumentConverter.java
```diff
@@ -105,54 +113,86 @@ public BsonDocument fromRow(final InternalRow row) {
    * @return a BsonDocument representing the data in the row
    */
   public BsonDocument fromRow(final Row row) {
-    if (row.schema() == null) {
-      throw new DataException("Cannot convert Row without schema");
-    }
-    return toBsonValue(row.schema(), row).asDocument();
+    return rowToBsonDocument(row, mappers, ignoreNulls);
```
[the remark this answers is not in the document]
rozza: Done

src/main/java/com/mongodb/spark/sql/connector/schema/RowToBsonDocumentConverter.java
rozza: By using them in Spark (via the integration tests) they are tested - (Which lead me to create the interfaces for functions in the first place).

`StructField` is a Scala case class and is serializable.
`ConvertJson` is an enum so is inherently serializable.

src/main/java/com/mongodb/spark/sql/connector/read/MongoScanBuilder.java
```diff
@@ -306,7 +305,9 @@ private Optional<BsonValue> getBsonValue(final String fieldName, final Object va
           localSchema = (StructType) localField.dataType();
         }
       }
-      return Optional.of(CONVERTER.toBsonValue(localDataType, value));
+      Function<Object, BsonValue> objectToBsonValue =
+          createObjectToBsonValue(localDataType, WriteConfig.ConvertJson.FALSE, false);
+      return Optional.of(objectToBsonValue.apply(value));
```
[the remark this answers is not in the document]
rozza: Here the cost of the generating the filters is based on the number of filters applied to the `DataFrameReader`. In the case where non are applied there is no cost. This could be updated to cache the function on demand, but it would only be beneficial if the same fieldname had multiple filters applied to it.

As this is a single cost once per creation of the `DataFrameReader` its performance impact is much less than the purpose of this PR which changes the `DataFrameWriter` cost from: calculate for every `Row` to once per `DataFrameWriter`.

So until we have data saying this is a bottleneck I'd rather keep the code simpler.

rozza (commented):
Updated

jyemin (approved):
LGTM

src/main/java/com/mongodb/spark/sql/connector/schema/RowToBsonDocumentConverter.java
stIncMale: Terrible, I forgot that all enums are of the `Serializable` type.

src/main/java/com/mongodb/spark/sql/connector/read/MongoScanBuilder.java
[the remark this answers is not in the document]
stIncMale: Ross, thank you for explaining.

stIncMale (approved):
I approved the changes, but the following [question](https://github.com/mongodb/mongo-spark/pull/110#pullrequestreview-1909554959) still stands:

> Have you confirmed with manual experiments (as far as I know, we don't have anything automated) that the change has positive impact on the performance characteristics?

rozza:
Thanks @stIncMale have confirmed the performance improvements. This brings the performance back to v3.x versions of the connector.

More info in the HELP-55837 ticket and there is a script in the SPARK-419 ticket that was used for checking performance.

PR #111, opened 2024-03-22
Fix MongoClientCache for factories with same/equal configuration

zaza:
MongoDriverInformation does not implement hashCode/equals which effectively prevents DefaultMongoClientFactory from being used as a key in MongoClientCache i.e. even for the same configuration factories are never equal and create multiple, redundant cache entries.


zaza:
Please ignore, included more commits than intended.

PR #112, opened 2024-03-22
Fix MongoClientCache for factories with same/equal configurations

zaza:
MongoDriverInformation does not implement hashCode/equals which effectively prevents DefaultMongoClientFactory from being used as a key in MongoClientCache i.e. even for the same configuration factories are never equal and create multiple, redundant cache entries.


stp-pv:
I would also consider implementing equals() and hashCode() for SimpleMongoConfig to avoid surprises.

zaza:
> I would also consider implementing equals() and hashCode() for SimpleMongoConfig to avoid surprises.

I did that [here](https://github.com/mongodb/mongo-spark/pull/113) but can join both PRs it that makes sense.

zaza:
@rozza would joining both PRs help with getting this reviewed and (hopefully) merged soon?

rozza:
Thank you @zaza this has been merged into main 👍

PR #113, opened 2024-03-22
SimpleMongoConfig does not implement hashCode/equals

zaza:
Without equals the SimpleMongoConfigs cannot be used with DefaultMongoClientFactory as they break the equal check there.


rozza:
Thank you @zaza this has been merged into main

PR #114, opened 2024-04-09
Provide Spark 3.1 - 3.5 support

rozza:
Two core fixes:

1) Support the interface change in `SupportsNamespaces` which added `cascade` to `dropNamespace`.
2) Support change in API for creating serializers to and from `InternalRow` to `Row`

SPARK-413


.evergreen/config.yml
```diff
@@ -321,23 +321,40 @@ post:
   - func: "upload test results"
   - func: "cleanup"
 
+variables:
```
rozza, line 324: variables ftw

src/main/java/com/mongodb/spark/sql/connector/MongoCatalog.java
```diff
@@ -178,12 +178,11 @@ public void alterNamespace(final String[] namespace, final NamespaceChange... ch
...
   }
 
-  /**
-   * Drop a database.
-   *
-   * @param namespace (database) a multi-part namespace
-   * @return true if the namespace (database) was dropped
-   */
+  @Override
+  public boolean dropNamespace(final String[] namespace, final boolean cascade) {
```
rozza, line 182: `dropNamespace(namespace)` was removed in 3.4
`dropNamespace(namespace, cascade)` was added in 3.4 - probably with a default value for cascade.

src/main/java/com/mongodb/spark/sql/connector/SupportsNamespacesAdapter.java
```diff
@@ -0,0 +1,40 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+package com.mongodb.spark.sql.connector;
+
+/**
+ * An adapter for SupportsNamespaces as the API for dropNamespace changed between versions
+ */
+interface SupportsNamespacesAdapter {
```
rozza, line 22: Declare both `dropNamespace` methods then both can be overridden

src/main/java/com/mongodb/spark/sql/connector/schema/SchemaToExpressionEncoderFunction.java
```diff
@@ -0,0 +1,112 @@
...
+  private static final SerializableFunction<StructType, Seq<Attribute>> ATTRIBUTE_FUNCTION;
+
+  static {
+    Class<?> clazz;
+    Object moduleStatic = null;
+    Method applyMethod = null;
+    SerializableFunction<StructType, Seq<Attribute>> attributeFunction = null;
+
+    try {
+      clazz = Class.forName("org.apache.spark.sql.catalyst.encoders.RowEncoder$");
```
rozza, line 45: Spark 3.1-3.4 support

src/main/java/com/mongodb/spark/sql/connector/schema/SchemaToExpressionEncoderFunction.java
```diff
@@ -0,0 +1,112 @@
...
+        // Ignore - not available in this version of Spark
+      }
+    }
+
+    if (applyMethod == null || attributeFunction == null) {
+      String sparkVersion = org.apache.spark.package$.MODULE$.SPARK_VERSION_SHORT();
+      throw new ExceptionInInitializerError("Unsupported version of Spark: " + sparkVersion);
+    }
+
+    MODULE_STATIC = moduleStatic;
```
rozza, line 82: Cache the methods that we looked up.

src/main/java/com/mongodb/spark/sql/connector/schema/SchemaToExpressionEncoderFunction.java
```diff
@@ -0,0 +1,112 @@
...
+    APPLY_METHOD = applyMethod;
+    ATTRIBUTE_FUNCTION = attributeFunction;
+  }
+
+  @Override
+  public ExpressionEncoder<Row> apply(final StructType schema) {
+    ExpressionEncoder<Row> rowEncoder = MethodInvoker.invoke(APPLY_METHOD, MODULE_STATIC, schema);
+    return rowEncoder.resolveAndBind(ATTRIBUTE_FUNCTION.apply(schema), SimpleAnalyzer$.MODULE$);
```
rozza, line 90: Get the resolved ExpressionEncoder which can create valid serializers and deserializers

src/main/java/com/mongodb/spark/sql/connector/schema/ConverterHelper.java
```diff
@@ -31,6 +31,8 @@
 
 /** Shared converter helper methods and statics */
 public final class ConverterHelper {
+  static final SchemaToExpressionEncoderFunction SCHEMA_TO_EXPRESSION_ENCODER_FUNCTION =
```
rozza, line 34: Make static the function that is used to convert a schema to an expression encoder.

rozza:
@jyemin & @stIncMale this is now ready for review :)

.evergreen/config.yml
```diff
@@ -411,71 +427,69 @@ axes:
...
     display_name: "Integration tests: ${scala} ${spark} ${version} ${topology}"
     tags: ["integration-test"]
-    run_on:
-      - "ubuntu1804-test"
     tasks:
       - name: "integration-test-task"
 
   - matrix_name: "integration-tests-2-13"
-    matrix_spec: { scala: "2.13", spark: "3.2.2", version: "*", topology: "*"}
+    matrix_spec: { scala: "2.13", spark: "*", version: "*", topology: "*"}
```
jyemin, line 477: What's the rationale for the variants for which we run integration tests?  For instance, I'm unclear why we run integration tests with Spark 3.1, 3.2, and 3.5, but not 3.3 and 3.4.   Or why we only run tests against latest sharded clusters but all MongoDB releases for replica sets.

.evergreen/config.yml
```diff
@@ -411,71 +427,69 @@ axes:
...
-      - "ubuntu1804-test"
+    <<: *exclude-spec
     tags: ["static-check"]
     tasks:
       - name: "static-checks-task"
 
   - matrix_name: "unit-tests"
     matrix_spec: { scala: "*", spark: "*" }
-    exclude_spec: {scala: "2.13", spark: "3.1.2"}
+    <<: *exclude-spec
```
jyemin, line 463: Should static checks and unit tests run on all spark versions? Seems odd that the results would be different (especially for the static checks)

build.gradle.kts
```diff
@@ -49,9 +49,9 @@ repositories {
     mavenCentral()
 }
 
-// Usage: ./gradlew -DscalaVersion=2.12 -DsparkVersion=3.1.2
+// Usage: ./gradlew -DscalaVersion=2.12 -DsparkVersion=3.1.4
 val scalaVersion = System.getProperty("scalaVersion", "2.13")
-val sparkVersion = System.getProperty("sparkVersion", "3.2.2")
+val sparkVersion = System.getProperty("sparkVersion", "3.2.4")
```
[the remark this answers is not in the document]
rozza: This is just the default when we run locally and can be changed as needed (eg its set in evergreen).

Probably should default to the latest version though - so will update.

src/main/java/com/mongodb/spark/sql/connector/MongoCatalog.java
rozza: Scala allows default values for parameters - so they may have just done that.

.evergreen/config.yml
rozza: The exclude spec definition has the info:

> # Scala 2.12 only works with 3.1.3 spark
> # Scala 2.13 doesn't work on 3.1.3 spark

So we exclude versions that would not compile

.evergreen/config.yml
rozza: I've updated this as it wasn't clear. The rationale should be to reduce the matrix to cover the requirements.

Scala 2.12 is tested against the only Spark version that supports it.
Scala 2.13 is run against the minmum and highest supported versions of Spark and the minimum and highest versions of MongoDB.  

The latest version of MongoDB is also included to act as a canary.

.evergreen/config.yml
jyemin: I'm asking the opposite question, actually.  I understand the exclusions, but I'm not clear why we would need to run static checks more than once (e.g. on Java driver we run static-analysis tests only with Java 17, not 8, 11 and 17).

stIncMale (changes requested):

.evergreen/config.yml
rozza: Because different versions of Scala run different code paths (there is separate code for Scalal 2.12 and 2.13 in the connector. 

There are also differences in versions of Spark that we compile against.  Eg Spark 3.4 added the `dropDatabase` change and Spark 3.5 added the encoder changes.  

The unit tests and static checks cover those code paths.

.evergreen/run-sharded-integration-tests.sh
```diff
@@ -0,0 +1,24 @@
...
+set -o xtrace   # Write all commands first to stderr
+set -o errexit  # Exit the script with error if any of the commands fail
+
+# Supported/used environment variables:
+# MONGODB_URI     Set the suggested connection MONGODB_URI (including credentials and topology info)
+# SCALA_VERSION   The Scala version to compile with
+# SPARK_VERSION   The spark version to test against
+MONGODB_URI=${MONGODB_URI:-}
+SCALA_VERSION=${SCALA_VERSION:-2.12}
+SPARK_VERSION=${SPARK_VERSION:-3.1.2}
```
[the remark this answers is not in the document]
rozza: I'd rather keep the defaults at the lowest supported version rather than the highest.
Not that it makes a real difference as they are passed in.

.evergreen/config.yml
```diff
@@ -411,71 +438,72 @@ axes:
...
         variables:
-          SPARK: "3.1.2"
-      - id: "3.2.2"
-        display_name: "Spark 3.2.2"
+          SPARK: "3.4.2"
+      - id: "3.5.1"
+        display_name: "Spark 3.5.1"
         variables:
-          SPARK: "3.2.2"
+          SPARK: "3.5.1"
```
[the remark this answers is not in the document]
rozza: I don't believe so - we produce two artifacts:
 - Scala 2.12 version only supports Spark 3.1.2
 - Scala 2.13 version supports Spark 3.2 - 3.5

src/main/java/com/mongodb/spark/sql/connector/schema/SchemaToExpressionEncoderFunction.java
```diff
@@ -0,0 +1,120 @@
...
+import java.lang.reflect.Method;
+import org.apache.spark.sql.Row;
+import org.apache.spark.sql.catalyst.analysis.SimpleAnalyzer$;
+import org.apache.spark.sql.catalyst.encoders.ExpressionEncoder;
+import org.apache.spark.sql.catalyst.expressions.Attribute;
+import org.apache.spark.sql.types.StructType;
+import scala.collection.immutable.Seq;
+
+@SuppressWarnings("unchecked")
+class SchemaToExpressionEncoderFunction
```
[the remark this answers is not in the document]
rozza: Yeah, its no longer needed - removed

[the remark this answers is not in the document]
rozza: Done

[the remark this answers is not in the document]
stIncMale: I don't see any new changes to the PR, you probably haven't pushed them.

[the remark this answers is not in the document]
rozza: image

Done!

stIncMale (approved):

jyemin (approved):
LGTM

PR #116, opened 2024-04-12
Fix handling of empty strings when parsing

rozza:
SPARK-421

### DIFF
--- a/src/main/java/com/mongodb/spark/sql/connector/schema/RowToBsonDocumentConverter.java
+++ b/src/main/java/com/mongodb/spark/sql/connector/schema/RowToBsonDocumentConverter.java
@@ -274,8 +274,8 @@ public final class RowToBsonDocumentConverter implements Serializable {
   private static final char JSON_ARRAY_START = '[';
   private static final char JSON_ARRAY_END = ']';
 
-  private static boolean isJsonObjectOrArray(@Nullable final String data) {
-    if (data == null || data.isEmpty()) {
+  private static boolean isJsonObjectOrArray(final String data) {
+    if (data.isEmpty()) {
       return false;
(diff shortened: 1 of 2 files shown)


src/main/java/com/mongodb/spark/sql/connector/schema/RowToBsonDocumentConverter.java
```diff
@@ -273,7 +274,10 @@ private static BsonDocument rowToBsonDocument(
   private static final char JSON_ARRAY_START = '[';
   private static final char JSON_ARRAY_END = ']';
 
-  private static boolean isJsonObjectOrArray(final String data) {
+  private static boolean isJsonObjectOrArray(@Nullable final String data) {
+    if (data == null || data.isEmpty()) {
+      return false;
+    }
```
stIncMale, line 280: `data` can't currently be `null`, so there is no need to relax the method requirements my allowing `null` values, which our Spark connector don't currently pass anyway.

```suggestion
  private static boolean isJsonObjectOrArray(final String data) {
    if (data.isEmpty()) {
      return false;
    }
```

stIncMale (changes requested):

src/main/java/com/mongodb/spark/sql/connector/schema/RowToBsonDocumentConverter.java
stIncMale: The `import org.jetbrains.annotations.Nullable;` line is also not needed anymore. It's surprising that the static checks don't complain about it.

I am approving to speedup the process: once the line is removed, feel free to merge.

stIncMale (approved):

PR #117, opened 2024-04-17
Fix convertJson configuration parsing

rozza:
SPARK-425

### DIFF
--- a/src/main/java/com/mongodb/spark/sql/connector/config/WriteConfig.java
+++ b/src/main/java/com/mongodb/spark/sql/connector/config/WriteConfig.java
@@ -25,7 +25,6 @@ import com.mongodb.WriteConcern;
 import com.mongodb.spark.sql.connector.exceptions.ConfigException;
 import java.util.HashMap;
 import java.util.List;
-import java.util.Locale;
 import java.util.Map;
 import java.util.concurrent.TimeUnit;
 import org.jetbrains.annotations.ApiStatus;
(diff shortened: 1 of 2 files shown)


src/main/java/com/mongodb/spark/sql/connector/config/WriteConfig.java
```diff
@@ -87,16 +86,28 @@ public enum ConvertJson {
       this.value = operationType;
     }
 
-    static ConvertJson fromString(final String jsonType) {
-      if (jsonType.equalsIgnoreCase(TRUE)) {
+    static ConvertJson fromString(final String operationType) {
```
stIncMale, line 89: [optional]
Not that I understand how the previous name "jsonType" was a good description for a value of the `convertJson` option, but neither do I understand why "operationType" is a good description for that. Wouldn't `operationType` or even just `v` be more relevant?

src/test/java/com/mongodb/spark/sql/connector/config/MongoConfigTest.java
```diff
@@ -127,6 +127,22 @@ void testMongoConfigWriteConcern() {
     assertEquals(new WriteConcern("region1"), writeConfig.getWriteConcern());
   }
 
+  @Test
+  void testWriteConfigConvertJson() {
+    WriteConfig writeConfig = MongoConfig.createConfig(CONFIG_MAP).toWriteConfig();
+    assertEquals(writeConfig.convertJson(), WriteConfig.ConvertJson.FALSE);
+    assertEquals(
+        writeConfig.withOption("convertJson", "True").convertJson(), WriteConfig.ConvertJson.ANY);
```
stIncMale, line 135: [optional]
Let's also add
```java
assertEquals(
        writeConfig.withOption("convertJson", "False").convertJson(), WriteConfig.ConvertJson.FALSE);
```

stIncMale (approved):
There are two optional items. Feel free to merge with or without addressing them.

src/main/java/com/mongodb/spark/sql/connector/config/WriteConfig.java
rozza: I just used it as its the same name in the constructor.  Will update.

src/test/java/com/mongodb/spark/sql/connector/config/MongoConfigTest.java
rozza: Done

PR #118, opened 2024-06-06
Added Schema hints for use when inferring schemas.

rozza:
Added a new configuration: `schemaHints`.
Users can now supply schema to enforce the schema information about known field types when inferring schema.

Supports the following Spark formats:
  - DDL: `value STRING,count INT`
  - SQL DDL: `STRUCT<value: STRING, count: INT>`
  - Simple String DDL: `struct<value:string,count:int>`
  - JSON: ```{"type":"struct","fields":[ {"name":"value","type":"string","nullable":true}, {"name":"count","type":"integer","nullable":true}]}```

To create DDL or Json schema strings simply use the Spark shell:

```
import org.apache.spark.sql.types._
val mySchema = StructType(Seq(StructField("value", StringType), StructField("count", IntegerType)))

mySchema.toDDL
mySchema.sql
mySchema.simpleString
mySchema.json
```

Or in PySpark:

```
from pyspark.sql.types import StructType, StructField, StringType, IntegerType
mySchema = StructType([ StructField('value',  StringType(), True), StructField('count', IntegerType(), True)])

mySchema.simpleString()
mySchema.json()
```

SPARK-365

### DIFF
--- a/src/main/java/com/mongodb/spark/sql/connector/schema/InferSchema.java
+++ b/src/main/java/com/mongodb/spark/sql/connector/schema/InferSchema.java
@@ -70,8 +70,6 @@ public final class InferSchema {
   /** Inferred schema metadata */
   public static final Metadata INFERRED_METADATA = Metadata.fromJson("{\"inferred\": true}");
 
-  private static final StructType EMPTY_SCHEMA = new StructType();
-
   /**
    * Infer the schema for the specified configuration.
    *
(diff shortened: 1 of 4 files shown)


src/main/java/com/mongodb/spark/sql/connector/schema/InferSchema.java
```diff
@@ -392,6 +446,52 @@ private static DataType dataTypeCheckStructTypeToMapType(
...
+        }
+      });
+
+      rhsMap.forEach((k, v) -> {
+        if (!newFields.containsKey(k)) {
+          newFields.put(k, v);
+        }
+      });
+
+      return new StructType(newFields.values().toArray(new StructField[0]));
```
vbabanin, line 488:
```suggestion
      // Iterate over rhs fields to merge or add.
      for (StructField rhsField : ((StructType) rhs).fields()) {
        String fieldName = rhsField.name();
        StructField lhsField = lhsMap.get(fieldName);

       // Merge field if exists in lhs; else, add from rhs.
        StructField newField = lhsField == null ? rhsField : new StructField(
                fieldName,
                rhsPreferredMerge(lhsField.dataType(), rhsField.dataType()),
                rhsField.nullable(),
                rhsField.metadata());

        lhsMap.put(fieldName, newField);
      }

      return new StructType(lhsMap.values().toArray(new StructField[0]));
```

It seems we might streamline the code by eliminating the creation of additional `rhsMap` and `newFields` maps. Specifically:

```
Map<String, StructField> rhsMap = Arrays.stream(((StructType) rhs).fields())
    .collect(Collectors.toMap(StructField::name, f -> f));
Map<String, StructField> newFields = new HashMap<>();
```

We could consider directly merging the `rhs` fields into the `lhs` map with overrides, since `lhs` is local and not shared elsewhere. Since `lhsMap` is local and not shared elsewhere, this would allow us to simply return it.

src/test/java/com/mongodb/spark/sql/connector/schema/InferSchemaTest.java
```diff
@@ -512,6 +512,142 @@ void testMapTypes() {
...
+            singletonList(BsonDocument.parse("{a: {b: {c: 'string', d: 'string'}}}")), readConfig));
+
+    // Ensure can handle multiple nested hints
+    readConfig = READ_CONFIG.withOption(ReadConfig.SCHEMA_HINTS, "`a.b.c` Int,`a.b.d` String");
+    assertEquals(
+        expected, InferSchema.inferSchema(singletonList(BsonDocument.parse("{}")), readConfig));
+  }
+
+  @Test
+  public void rhsPreferredMergeTest() {
```
vbabanin, line 613:
```suggestion
  void rhsPreferredMergeTest() {
```

src/main/java/com/mongodb/spark/sql/connector/schema/InferSchema.java
rozza: Done - renamed to clarify

vbabanin (approved):

vbabanin:
LGTM!

PR #119, opened 2024-06-12
Added parse mode support when reading data from MongoDB.

rozza:
Adds the `mode` configuration allowing for different parsing strategies when handling documents that don't match the expected schema during reads.

The options are:

  - `FAILFAST` (default) throw an exception when parsing a document that doesn't match the schema.
  - `PERMISSIVE` Sets any invalid fields to `null`. Combine with the `columnNameOfCorruptRecord` configuration if you want to store any invalid documents as an extended json string.
  - `DROPMALFORMED` ignores the whole document.

Adds the `columnNameOfCorruptRecord` configuration whic extends the `PERMISSIVE` mode. When configured it saves the whole invalid document as extended json in that column, as long as its defined in the Schema. Inferred schemas will add the `columnNameOfCorruptRecord` column if set and the `mode` is `PERMISSIVE`.

Note: Names derive from existing spark json configurations, from where this feature takes inspiration.

SPARK-327

### DIFF
--- a/src/integrationTest/java/com/mongodb/spark/sql/connector/read/AbstractMongoStreamTest.java
+++ b/src/integrationTest/java/com/mongodb/spark/sql/connector/read/AbstractMongoStreamTest.java
@@ -23,11 +23,15 @@ import static java.lang.String.join;
 import static java.util.Arrays.asList;
 import static java.util.Collections.emptyList;
 import static java.util.Collections.singletonList;
+import static java.util.Collections.unmodifiableList;
 import static java.util.stream.Collectors.joining;
 import static org.apache.spark.sql.types.DataTypes.createStructField;
 import static org.apache.spark.sql.types.DataTypes.createStructType;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertInstanceOf;
+import static org.junit.jupiter.api.Assertions.assertIterableEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.junit.jupiter.api.Assertions.fail;
@@ -38,7 +42,6 @@ import com.mongodb.client.MongoDatabase;
 import com.mongodb.client.model.Filters;
 import com.mongodb.client.model.InsertManyOptions;
 import com.mongodb.client.model.Updates;
-import com.mongodb.lang.Nullable;
 import com.mongodb.spark.sql.connector.config.CollectionsConfig;
 import com.mongodb.spark.sql.connector.config.MongoConfig;
 import com.mongodb.spark.sql.connector.config.ReadConfig;
@@ -66,12 +69,17 @@ import org.apache.spark.sql.SparkSession;
 import org.apache.spark.sql.streaming.DataStreamReader;
 import org.apache.spark.sql.streaming.StreamingQuery;
 import org.apache.spark.sql.streaming.StreamingQueryException;
+import org.apache.spark.sql.streaming.StreamingQueryListener;
 import org.apache.spark.sql.streaming.Trigger;
+import org.apache.spark.sql.types.DataType;
 import org.apache.spark.sql.types.DataTypes;
 import org.apache.spark.sql.types.StructType;
 import org.bson.BsonDocument;
+import org.bson.BsonInt32;
+import org.bson.BsonNull;
 import org.bson.BsonString;
 import org.bson.BsonTimestamp;
+import org.jetbrains.annotations.Nullable;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.ValueSource;
@@ -327,7 +335,7 @@ abstract class AbstractMongoStreamTest extends MongoSparkConnectorTestCase {
   }
 
   @ParameterizedTest
-  @ValueSource(strings = {"SINGLE", "MULTIPLE", "ALL"})
+  @ValueSource(strings = {"SINGLE", "MULTIPLE"})
   void testStreamResumable(final String collectionsConfigModeStr) {
     assumeTrue(supportsChangeStreams());
     assumeTrue(isAtLeastFourDotFour());
@@ -338,7 +346,7 @@ abstract class AbstractMongoStreamTest extends MongoSparkConnectorTestCase {
     MongoConfig mongoConfig = createMongoConfig(collectionsConfigType);
 
     testStreamingQuery(
-        "mongodb",
+        MONGODB,
         mongoConfig,
         withSource("inserting 0-25", (msg, coll) -> coll.insertMany(createDocuments(0, 25))),
         withSink(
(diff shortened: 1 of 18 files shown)


src/integrationTest/java/com/mongodb/spark/sql/connector/read/AbstractMongoStreamTest.java
```diff
@@ -327,7 +335,7 @@ void testStreamPublishFullDocumentOnlyHandlesCollectionDrop(
   }
 
   @ParameterizedTest
-  @ValueSource(strings = {"SINGLE", "MULTIPLE", "ALL"})
+  @ValueSource(strings = {"SINGLE", "MULTIPLE"})
```
rozza, line 338: This fixes a bug where it was checking the writes against the source db and not the sink db!  As you cannot test reading from all collections and writing to that db at the same time. I opted to remove the ALL step. Added SPARK-429 to track.

src/integrationTest/java/com/mongodb/spark/sql/connector/read/AbstractMongoStreamTest.java
```diff
@@ -598,7 +721,7 @@ private final void testStreamingQuery(
       final MongoConfig mongoConfig,
       final Consumer<MongoConfig> setup,
       final Consumer<MongoConfig>... consumers) {
-    testStreamingQuery(writeFormat, mongoConfig, DEFAULT_SCHEMA, null, setup, consumers);
+    testStreamingQuery(writeFormat, mongoConfig, DEFAULT_SCHEMA, null, null, setup, consumers);
```
rozza, line 724: The changes below are test infrastructure related eg adding stream listener support and / or simplify the testing.

src/integrationTest/java/com/mongodb/spark/sql/connector/read/AbstractMongoStreamTest.java
```diff
@@ -765,7 +920,7 @@ private Consumer<MongoConfig> withSink(
       WriteConfig writeConfig = mongoConfig.toWriteConfig();
       MongoCollection<BsonDocument> collection = writeConfig.withClient(client -> client
           .getDatabase(writeConfig.getDatabaseName())
-          .getCollection(collectionName(), BsonDocument.class));
+          .getCollection(writeConfig.getCollectionName(), BsonDocument.class));
```
rozza, line 923: Use the write config to source the collection name to write to!!

src/integrationTest/java/com/mongodb/spark/sql/connector/read/AbstractMongoStreamTest.java
```diff
@@ -799,6 +954,9 @@ private MongoConfig createMongoConfig(final CollectionsConfig.Type collectionsCo
     }
     options.put(
         ReadConfig.READ_PREFIX + ReadConfig.COLLECTION_NAME_CONFIG, collectionsConfigOptionValue);
+    options.put(
```
rozza, line 957: Sets the write config collection name.

src/main/java/com/mongodb/spark/sql/connector/config/ReadConfig.java
```diff
@@ -161,6 +161,66 @@ public final class ReadConfig extends AbstractMongoConfig {
...
+     * Ignore any data errors for a field and replace with null
+     */
+    PERMISSIVE,
+    /**
+     * Drop any rows that contains a data error
+     */
+    DROPMALFORMED;
+
+    static ParseMode fromString(final String userParseMode) {
+      try {
```
vbabanin, line 182: If `null` is provided through the `option()` method, it currently results in a `NullPointerException`. To make this more user-friendly and consistent, we could throw a `ConfigException` , which provides clearer feedback.

```suggestion
     validateConfig(userParseMode, Objects::nonNull, () -> "The userParseMode can't be null");
      try {
      
```

src/main/java/com/mongodb/spark/sql/connector/schema/BsonDocumentToRowConverter.java
```diff
@@ -110,12 +117,51 @@ public StructType getSchema() {
...
+          try {
+            valueMap.put(
+                field.name(),
+                convertBsonValue(field.name(), field.dataType(), bsonDocument.get(field.name())));
+          } catch (DataException e) {
+            hasError = true;
+            if (!isPermissive) {
+              throw e;
+            }
+            valueMap.put(field.name(), null);
```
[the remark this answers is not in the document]
rozza: Great question. I initially thought this wasn't defined and in a closed source module. However, having searched the spark open source code it is native to their json importer. So I checked the behavior in pyspark and it appears it is recursive!

```
>>> from pyspark.sql.types import _parse_datatype_string
>>> schema = _parse_datatype_string('struct<a:string,b:struct<c:string,d:bigint>,_corrupt_record:string>')
>>>
>>> json = ['{"a":"ok","b":{"c":"c","d":10}}','{"a":"bad","b":{"c":"c","d":"bad"}}']
>>> stringdf = sc.parallelize(json)
>>> df = spark.read.option('mode', 'PERMISSIVE').schema(schema).json(stringdf)
>>> df.show()
+---+---------+--------------------+
|  a|        b|     _corrupt_record|
+---+---------+--------------------+
| ok|  {c, 10}|                NULL|
|bad|{c, NULL}|{"a":"bad","b":{"...|
+---+---------+--------------------+
```

I will update :)

src/main/java/com/mongodb/spark/sql/connector/config/ReadConfig.java
rozza: I ended up reverting as this method is internal and only called via the readConfig method where there is a default value for the parse method.

src/main/java/com/mongodb/spark/sql/connector/schema/BsonDocumentToRowConverter.java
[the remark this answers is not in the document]
rozza: Done and it was simplier than expected to add 👍

src/main/java/com/mongodb/spark/sql/connector/schema/BsonDocumentToRowConverter.java
```diff
@@ -212,7 +243,7 @@ private GenericRowWithSchema convertToRow(
         .forEach((k, v) ->
             map.put(k, convertBsonValue(createFieldPath(fieldName, k), dataType.valueType(), v)));
 
-    return JavaScala.asScala(map);
+    return scala.collection.immutable.Map$.MODULE$.from(JavaScala.asScala(map));
```
rozza, line 246: Found this bug when converting a row to json and Spark errored due to it not being the correct (immutable) scala map type.

src/main/java/com/mongodb/spark/sql/connector/schema/BsonDocumentToRowConverter.java
vbabanin, line 1: In PERMISSIVE mode, I see that corrupted fields are set to null, regardless of the schema specifications. 

However, in `covertToRow()` method when non-nullable field defined in schema is missing in the document, we throw an exception:` com.mongodb.spark.sql.connector.exceptions.DataException: Missing field 'fieldName'`

```
   if (hasField || field.nullable()) {
        values.add(
            hasField
                ? convertBsonValue(fullFieldPath, field.dataType(), bsonDocument.get(field.name()))
                : null);
      } else {
        throw missingFieldException(fullFieldPath, bsonDocument);
      }  
```
Given the unstructured nature of our DB data, where documents may vary in field counts, should we consider setting absent fields expected by the schema as null when operating in PERMISSIVE mode? I'm concerned that throwing exceptions in these situations could create issues for some users.

In similar contexts, Spark's strategy with CSV files sets missing fields to null when a record has fewer tokens than the schema expects. Furthermore, according to the [Spark Avro documentation](https://spark.apache.org/docs/3.5.1/sql-data-sources-avro.html#data-source-option), in PERMISSIVE mode: 'Corrupt records are processed as null result. Therefore, the data schema is forced to be fully nullable, which might be different from the one user provided,' implying a schema that is forced to full nullability.

Additionally, we might consider introducing another mode that allows handling missing fields without triggering failures. This approach could provide users with more flexibility in managing data discrepancies

rozza: Done, now force nullability. This is inline with json parsing in Spark.

vbabanin (approved):
LGTM!

PR #124, opened 2024-08-14
SPARK-414 support for delete

VVBondarenko:
- added support for delete
- minor refactoring for filter transformations
- minor fix for catalog to load table with resolved schema instead of empty one


rozza:
Closed in favor of: #144 - which takes this PR as a base

PR #127, opened 2024-11-27
Allow reuse of MongoClient instances for the same collection

spingel:
Spark pushes down aggregation pipelines in the configuration when acquiring MongoClient instances. This change modifies the MongoClientFactory to only take the connection into account for lookups in MongoClientCache.

Without this change different queries against the same MongoDB database result in multiple MongoClient instances preventing caching in those scenarios.

SPARK-441


rozza:
Thanks @spingel - this will be released in 10.4.1 which is scheduled for release Feb 4th

PR #128, opened 2024-11-27
Prefer mongodb.keep_alive_ms over spark.mongodb.keep_alive_ms property

spingel:
The documentation at https://www.mongodb.com/docs/spark-connector/master/faq/ specifies to set mongodb.keep_alive_ms but in the code the system property is prefixed with spark. This changes makes the code consistent with the documentation while maintaining backwards compatibility.

When attempting to set this with
--conf spark.executor.extraJavaOptions="-Dspark.mongodb.keep_alive_ms=70000" an exception occurs since Spark does not allow setting spark flags with that mechanism on executors:

 java.lang.Exception: spark.executor.extraJavaOptions is not allowed to set Spark
 options (was '-Dspark.mongodb.keep_alive_ms=70000’). Set them directly on a
 SparkConf or in a properties file when using ./bin/spark-submit.

Work-around: Rather than setting the system property via extraJavaOptions on executors, the environment variable JAVA_TOOL_OPTIONS can be used:

--conf spark.driver.extraJavaOptions=-Dspark.mongodb.keep_alive_ms=70000 --conf spark.executorEnv.JAVA_TOOL_OPTIONS=-Dspark.mongodb.keep_alive_ms=70000

### DIFF
--- a/src/main/java/com/mongodb/spark/sql/connector/connection/LazyMongoClientCache.java
+++ b/src/main/java/com/mongodb/spark/sql/connector/connection/LazyMongoClientCache.java
@@ -31,19 +31,35 @@ public final class LazyMongoClientCache {
 
   private static final MongoClientCache CLIENT_CACHE;
 
-  private static final String SYSTEM_MONGO_CACHE_KEEP_ALIVE_MS_PROPERTY =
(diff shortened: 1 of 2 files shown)


rozza:
Thanks @spingel this was rebased and merged in [6fdb33d](https://github.com/mongodb/mongo-spark/commit/6fdb33d38d634bd816cc31b3edfc9c2b0cded983)

PR #129, opened 2024-11-27
Update tests in MongoClientCacheTest

zaza:
* `factoriesWithEqualSimpleConfigsCreateSameClientsThroughCache` can be now enabled since the [referenced PR](https://github.com/mongodb/mongo-spark/pull/112) has been merged, without that fix it would fail
* `factoriesWithSameSimpleConfigCreateSameClientsThroughCache` can be removed as it is identical to `factoriesWithSameConfigCreateSameClientsThroughCache`


rozza:
Thanks @zaza

PR #130, opened 2025-01-15
Improved support for MongoOffsets using hadoop file systems

rozza:
Spark expects hadoop configurations to be prefixed with `spark.hadoop`. However, documentation on the web omits this prefix when setting filesystem configuration - See the azure storage docs.

The issue the connector has is the `MongoOffset` support just uses the `SparkContext.hadoopConfiguration()` helper method, which omits any non-prefixed configuration. So this improvement adds any filesystem configuration prefixed with `fs.` to the hadoop configuration. This ensures that the `MongoOffset`s use of the Hadoop filesystem includes the configuraion.

SPARK-438


katcharov (approved):
LGTM!

PR #131, opened 2025-01-22
Bump minimum MongoDB version tested to MongoDB 4.4

rozza:
SPARK-439


rozza:
4.4 is the minimum tested in the Java driver and 4.0 seems no longer available in ubuntu 2004 via evergreen (https://github.com/alcaeus/drivers-evergreen-tools/blob/master/.evergreen/download-mongodb.sh#L321-L327)

nhachicha (approved):

PR #132, opened 2025-02-14
Use computed average document size to support usage through Atlas Data Federation

guillotjulien:
The new implementation stops relying on the storageStats property that is not being recognized as a valid property when using the $collStats aggregation operation with a Data Federation endpoint.
This end up making it impossible to use the SamplePartitioner, PaginateBySizePartitioner and AutoBucketPartitioner in that situation.

Example:
```
Caused by: com.mongodb.MongoCommandException: Command failed with error 9 (FailedToParse): '$collStats param 'storageStats' is not valid for Atlas Data Federation, correlationID = 182415f1c5629184818f0150' on server <REDACTED>. The full response is {"ok": 0, "errmsg": "$collStats param 'storageStats' is not valid for Atlas Data Federation, correlationID = 182415f1c5629184818f0150", "code": 9, "codeName": "FailedToParse"}
        at com.mongodb.internal.connection.ProtocolHelper.getCommandFailureException(ProtocolHelper.java:198)
        at com.mongodb.internal.connection.InternalStreamConnection.receiveCommandMessageResponse(InternalStreamConnection.java:416)
        at com.mongodb.internal.connection.InternalStreamConnection.sendAndReceive(InternalStreamConnection.java:340)
        at com.mongodb.internal.connection.UsageTrackingInternalConnection.sendAndReceive(UsageTrackingInternalConnection.java:116)
        at com.mongodb.internal.connection.DefaultConnectionPool$PooledConnection.sendAndReceive(DefaultConnectionPool.java:643)
        at com.mongodb.internal.connection.CommandProtocolImpl.execute(CommandProtocolImpl.java:71)
        at com.mongodb.internal.connection.DefaultServer$DefaultServerProtocolExecutor.execute(DefaultServer.java:206)
        at com.mongodb.internal.connection.DefaultServerConnection.executeProtocol(DefaultServerConnection.java:119)
        at com.mongodb.internal.connection.DefaultServerConnection.command(DefaultServerConnection.java:85)
        at com.mongodb.internal.connection.DefaultServerConnection.command(DefaultServerConnection.java:75)
        at com.mongodb.internal.connection.DefaultServer$OperationCountTrackingConnection.command(DefaultServer.java:293)
        at com.mongodb.internal.operation.CommandOperationHelper.createReadCommandAndExecute(CommandOperationHelper.java:233)
        at com.mongodb.internal.operation.CommandOperationHelper.lambda$executeRetryableRead$4(CommandOperationHelper.java:215)
[...]

PR #133, opened 2025-04-24
Spark 442

rozza:
Simplified approach to #132

Ensures the whole collection count is used when calculating the average size of documents.

### DIFF
--- a/src/main/java/com/mongodb/spark/sql/connector/read/partitioner/PartitionerHelper.java
+++ b/src/main/java/com/mongodb/spark/sql/connector/read/partitioner/PartitionerHelper.java
@@ -19,7 +19,6 @@ package com.mongodb.spark.sql.connector.read.partitioner;
 
 import static com.mongodb.spark.sql.connector.read.partitioner.Partitioner.LOGGER;
 import static java.util.Arrays.asList;
-import static java.util.Collections.singletonList;
 
 import com.mongodb.MongoCommandException;
 import com.mongodb.client.MongoDatabase;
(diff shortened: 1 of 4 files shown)


rozza:
@guillotjulien please check out this alternative to #132

src/main/java/com/mongodb/spark/sql/connector/read/partitioner/PartitionerHelper.java
```diff
@@ -101,21 +107,43 @@ public static List<BsonDocument> createPartitionPipeline(
...
+      return readConfig.withCollection(
+          coll -> Optional.ofNullable(coll.aggregate(COLL_STATS_DATA_FEDERATION_AGGREGATION_PIPELINE)
+                  .allowDiskUse(readConfig.getAggregationAllowDiskUse())
+                  .comment(readConfig.getComment())
+                  .map(collStats -> collStats.append(
+                      "size",
+                      collStats
+                          .getDocument("partition", new BsonDocument())
+                          .getNumber("size", new BsonInt32(0))))
+                  .first())
```
guillotjulien, line 144: One issue with that approach is that when you have multiple partitions (S3 or Mongo clusters) backing up your federated collection, you'll only use the info from the first partition. Whereas the `collStats` command gives you the total size of all partitions.

I think this can be mitigated by summing the partition size and count of all partitions, then returning a `BsonDocument` from that.

rozza: Thanks @guillotjulien - according to the [atlas docs](https://www.mongodb.com/docs/atlas/data-federation/supported-unsupported/pipeline/collstats/)

> The following example shows $collStats syntax for retrieving the total number of documents in the partitions.
> ```
> use s3Db
> db.abc.aggregate([ {$collStats: {"count" : {} }} ])
> ```

Have you found that is not the case?

guillotjulien: @rozza it does indeed, but it gives you one document per partition as shown in the example above. So if you want the total, you'd need to compute the sum of counts of all partitions.

e.g.
```
[
  { $collStats: { "count" : {} } },
  { $group: { _id: null, totalCount: { $sum: "$count" }, totalSize: { $sum: "$partition.size" } }  }
]
```

rozza: Thanks @guillotjulien I've added that to the latest commit.

guillotjulien:
Hi @rozza, I checked locally, and everything is working as expected. So LGTM thanks!

rozza:
@guillotjulien just to let you know 10.5.0 has been released and is on Maven central.

guillotjulien:
Thanks for the fast release @rozza!

PR #134, opened 2025-04-30
Write config - default to the connection string writeconcern is avail…

rozza:
…able.

SPARK-392

### DIFF
--- a/src/main/java/com/mongodb/spark/sql/connector/config/WriteConfig.java
+++ b/src/main/java/com/mongodb/spark/sql/connector/config/WriteConfig.java
@@ -344,7 +344,10 @@ public final class WriteConfig extends AbstractMongoConfig {
   }
 
   private WriteConcern createWriteConcern() {
-    WriteConcern writeConcern = WriteConcern.ACKNOWLEDGED;
+    WriteConcern writeConcern = getConnectionString().getWriteConcern();
+    if (writeConcern == null) {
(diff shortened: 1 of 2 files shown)


src/main/java/com/mongodb/spark/sql/connector/config/WriteConfig.java
```diff
@@ -344,7 +344,10 @@ private String getNamespaceDescription(final CollectionsConfig collectionsConfig
   }
 
   private WriteConcern createWriteConcern() {
-    WriteConcern writeConcern = WriteConcern.ACKNOWLEDGED;
+    WriteConcern writeConcern = getConnectionString().getWriteConcern();
+    if (writeConcern == null) {
+      writeConcern = WriteConcern.ACKNOWLEDGED;
```
stIncMale, line 349: [optional]

To make it obvious for a code reader that the default write concern is the same as that in the `MongoClientSettings`, we could write `MongoClientSettings.builder().build().getWriteConcern()` instead of `WriteConcern.ACKNOWLEDGED`.

We may further document on `WriteConfig.WRITE_CONCERN_W_CONFIG` that the default write concern is the same as the one used by `MongoClientSettings`, which is `WriteConcern.ACKNOWLEDGED`.

stIncMale (approved):
There is one optional suggestion, feel free to ignore.

PR #140, opened 2025-07-15
Streaming added full document before change support

mrjoe7:
**Motivation**

We’re adding `change.stream.lookup.full.document.before.change` to ensure that we can access the full document prior to a delete operation in MongoDB change streams.

This is particularly important for our use case where:

- Deleted documents must be preserved for downstream consumers.
- The delete change event by default does not include the document body — without enabling pre-images, we would lose all context about what was removed.
- Having the pre-image ensures that we can reconstruct the full lifecycle of a document (create → update → delete) from the change stream alone.

We’re setting this as a configurable option to remain compatible with MongoDB deployments.


rozza:
Hi @mrjoe7,

Thanks for the PR - this looks like a great addition to the connector.  I've added [SPARK-449](https://jira.mongodb.org/browse/SPARK-449) to Jira to get this triaged and plan work.

I'm not sure we have any time planned this quarter for a release but will try to get this added asap.

Ross

PR #141, opened 2025-10-01
Update to Spark 4

rozza:
Bump version to 11.0.0 as its a major change
Updates to Spark 4 support
Removes Scala 2.12 support (not supported in Spark 4) 
Updates gradlew to 9.0.0
Updates to Evergreen configuration

JAVA-450


build.gradle.kts
```diff
@@ -121,22 +111,29 @@ tasks.withType<JavaCompile> {
...
-    }
-    describeStdOut.toString().replaceIndent(" - ")
+    providers
+        .exec {
+            isIgnoreExitValue = true
+            commandLine("git", "diff", "--name-only")
+        }
+        .standardOutput
+        .asText
+        .map { it.trim().replaceIndent("-") }
```
nhachicha: Is this needed? I think only if  the output has indentation, which I don't think git output has

build.gradle.kts
```diff
@@ -97,22 +92,17 @@ dependencies {
...
 
-val defaultJdkVersion: Int = 11
-
-java {
-    toolchain.languageVersion.set(JavaLanguageVersion.of(defaultJdkVersion))
-}
-
 tasks.withType<JavaCompile> {
     options.encoding = "UTF-8"
     options.release.set(8)
```
nhachicha: since we bumped to JDK 17 is it worth enabling `options.compilerArgs.add("-Werror")` to detect any deprecation ? (we're still releasing with byte code compatible with Java 8)

.evergreen/config.yml
```diff
@@ -412,76 +403,41 @@ axes:
...
+    matrix_spec: { version: ["4.4", "8.0", "latest"], topology: "replicaset" }
+    display_name: "Integration tests: ${version} ${topology}"
     tags: ["integration-test"]
     tasks:
       - name: "integration-test-task"
 
   - matrix_name: "integration-tests-2-13-sharded"
-    matrix_spec: { scala: "2.13", spark: [ "3.2.4", "3.5.1" ], version: ["4.4", "7.0", "latest" ], topology: "sharded" }
+    matrix_spec: { version: ["4.4", "8.0", "latest" ], spark: [ "4.0.1", "4.1.0" ], topology: "sharded" }
     display_name: "Integration tests: ${scala} ${spark} ${version} ${topology}"
```
nhachicha: The referenced  `${scala}` variable used in the display name has been removed above

nhachicha (commented):
LGTM 👍 minor comments.

- Do we have a target for testing on JDK8? (for runtime compatibility since we bumped to JDK17)
- [Doc] Is there a migration guide for breaking changes? Should we mention that there's no backward compatibility path for Spark 3.x users

build.gradle.kts
```diff
@@ -25,34 +27,28 @@ buildscript {
 
 plugins {
     idea
-    `java-library`
```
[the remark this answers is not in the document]
rozza: We just use the Java plugin

.evergreen/publish.sh
```diff
@@ -29,5 +29,4 @@ fi
 SYSTEM_PROPERTIES="-Dorg.gradle.internal.publish.checksums.insecure=true"
 
 ./gradlew -version
-./gradlew ${SYSTEM_PROPERTIES} --stacktrace --info ${TASK} -DscalaVersion=2.12 -DsparkVersion=3.1.2
-./gradlew ${SYSTEM_PROPERTIES} --stacktrace --info ${TASK} -DscalaVersion=2.13 -DsparkVersion=3.2.2
+./gradlew ${SYSTEM_PROPERTIES} --stacktrace --info ${TASK}
```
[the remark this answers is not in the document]
rozza: We just publish one artifact which covers Spark 4.x and Scala 2.13

Previously we published for both Scala 2.12 and Scala 2.13.  The complexity Scala 2.13 support was only added in Spark 3.2.2 so we had to compile against that version.

PR #155, opened 2026-08-15
SPARK-466

strogiyotec:
SPARK-466 add a setting to shuffle the partition list
Otherwise it's sorted and later on ExecutorService will execute partitions against the same shard
 spark.mongodb.read.partitioner.options.shuffle - boolean , When true, shuffles the generated partitions so consecutive partitions are unlikely to belong to the same shard, spreading read load across shards.
spark.mongodb.read.partitioner.options.shuffle.seed - long , The seed used when shuffling partitions. Set it to get a reproducible shuffle. Only takes effect when shuffle is true. An invalid (non-long) value raises a MongoSparkException.


stIncMale:
I removed myself from the reviewers for the same reason as in https://github.com/mongodb/mongo-spark/pull/156#issuecomment-5781404442.

PR #156, opened 2026-08-15
SPARK-466 - #155

strogiyotec:
[SPARK-466](https://jira.mongodb.org/browse/SPARK-466) add a setting to shuffle the partition list
Otherwise it's sorted and later on ExecutorService will execute partitions against the same shard
spark.mongodb.read.partitioner.options.shuffle - boolean , When true, shuffles the generated partitions so consecutive partitions are unlikely to belong to the same shard, spreading read load across shards.
spark.mongodb.read.partitioner.options.shuffle.seed - long , The seed used when shuffling partitions. Set it to get a reproducible shuffle. Only takes effect when shuffle is true. An invalid (non-long) value raises a MongoSparkException.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -38,7 +38,7 @@ plugins {
     id("io.github.gradle-nexus.publish-plugin") version "2.0.0"
 }
 
-version = "11.1.0"
+version = "11.2.0-SNAPSHOT"
 group = "org.mongodb.spark"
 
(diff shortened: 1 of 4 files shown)


stIncMale:
I removed myself from the reviewers because I was added by a bot, and @rozza is listed as a reviewer in https://jira.mongodb.org/browse/SPARK-466 (also, it makes sense for Ross to review this PR).

